PR #10, opened 2018-04-16
docs: more robust typescript instructions

Guria:
**What**: documentation to extend jest.Matchers interface in typescript

**Why**: proposed type-casting is ugly

**How**: edited README

**Checklist**:

* [x] Documentation
* [ ] Tests (N/A)
* [x] Ready to be merged 

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,27 +163,19 @@ expect(getByTestId(container, 'delete-button')).not.toHaveClass('btn-link')
 
 ### Using with Typescript
 
-When you use custom Jest Matchers with Typescript, you will need to extend the
-type signature of `jest.Matchers<void>`, then cast the result of `expect`
-accordingly. Here's a handy usage example:
+When you use custom Jest Matchers with Typescript, you can extend `jest.Matchers`
+interface provided by `@types/jest` package by creating `.d.ts` file somewhere
+in your project with following content:
 
 ```typescript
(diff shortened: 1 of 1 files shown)


gnapse:
Thanks for the contribution!

I'm not too knowledgable in Typescript (pending to learn more of it) so I can't really check your code, but I trust it's ok. It does look better than the current instructions. I think I was subconsciously hoping there would be a better way than that, and apparently there is.

I'll wait until you add yourself to the contributors table and then I'll merge.

gnapse:
I went on and merged this. Feel free to add yourself to the contributors list in a separate PR.

gnapse:
@Guria I tried to follow the typescript definitions instructions as they are in the README now, and it does not work (at least according to vscode, which continues to complain that "Property 'toHaveTextContent' does not exist on type 'Matchers<void>'").

The instruction say that creating a `.d.ts` file in the project is enough. How does the typescript system gets to know about this file and process it? Shouldn't it either have some specific name, or else be imported or referenced somehow from a config file or something?

gnapse:
Follow-up question: isn't there a way to provide these type definitions with this library? Otherwise if we ever add new matchers or change the definitions of existing matchers, users of this library will have outdated definitions, and if they ever get to realize that, they'd have to update them manually. Isn't there a way to ease this for the users of this library?

PR #13, opened 2018-05-18
feat: toBeVisible custom matcher

### LINKED ISSUE #7: New toBeVisible custom matcher
As discussed in #4, this library needs a `.toBeVisible` custom matcher. It operates on an element, receives no extra arguments, and passes if the element is visible to the user.

```javascript
expect(element).toBeVisible()
```

## How?

This should be implemented by checking the element's [computed style](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle), and make sure that `display` is not `none`, and also `visibility` is not `hidden` or `collapsed`.

## Why?

Doing the equivalent assertion using vanilla DOM APIs is cumbersome:

```javascript
// Two assertions needed, both dealing with verbose APIs
expect(getComputedStyle(element).getPropertyValue('display')).not.toEqual('none')
expect(getComputedStyle(element).getPropertyValue('visibility')).not.toMatch(/(hidden|collapsed)/)
```

gnapse:
**What**:

A new `.toBeVisible` custom matcher, as discussed in #7 

Closes #7 

**Why**:

Because checking for this manually is not trivial.

**How**:

Using `getComputedStyle` on an element to check if it's not visible either via `display: none` or `visibility: hidden|collapse`. Also checks parent nodes all the way to the top, if any of them is not visible, the element is not visible as well.

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ to maintain.
   * [`toHaveAttribute`](#tohaveattribute)
   * [`toHaveClass`](#tohaveclass)
   * [`toHaveStyle`](#tohavestyle)
+  * [`toBeVisible`](#tobevisible)
 * [Inspiration](#inspiration)
 * [Other Solutions](#other-solutions)
 * [Guiding Principles](#guiding-principles)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -194,6 +195,34 @@ This also works with rules that are applied to the element via a class name for
...
+### `toBeVisible`
+
+This allows you to check if an element is currently visible to the user.
+
+An element is visible if **all** the following conditions are met:
+
+* it does not have its css property `display` set to `none`
+* it does not have its css property `visibility` set to either `hidden` or
+  `collapse`
+* its parent element is also visible (and so on up to the top of the DOM tree)
```
kentcdodds: What about `opacity: 0`? That's a pretty common way to make things invisible as well. Hmmm...

gnapse: Hmmmmm, yes. Good point 🤔

gnapse:
I think the feature I like the most about semantic-release is that it leaves this message above in PRs once they get released after being merged. So many times I find an issue in a library that I'm having trouble with, I find that it has been merged, but there's no way to know if it's just on master, or beta or rc release, or the latest stable release. It's frustrating.

kentcdodds:
Yeah! I love that feature too!

PR #16, opened 2018-06-07
Remove snapshots

gnapse:
**What**:

Removing snapshots of error messages in the tests that are expected to throw.

**Why**:

Because the terminal output stored I these snapshots is not the same in terminals where colors are not supported, such as in Travis-CI.

**Checklist**:

* [X] Tests
* [X] Ready to be merged


kentcdodds:
You could also strip the ansi characters with a snapshot serializer. I'm pretty sure there's a module that does this.

microcood:
Also disable chalk before tests ```beforeAll(() => {
  chalk.enabled = false;
});``` will work.
More interesting question why CI does not work. How its different between PR and master

gnapse:
There's a just issue I referenced in the PR that introduced this problem (facebook/jest#4407) where they talk a bit more about this, and that it is expected. They suggest the serializer there, but I haven't found any yet. Suggestions of where to look are welcome. I already looked in [jest-awesome](https://github.com/jest-community/awesome-jest) with no luck.

I also should've detected this in the original PR, because the codecov report never appeared and I did not notice.

BTW, interesting the `chalk.enabled` fix. I guess that's the real fallback I should use if the serializer does not appear. At least it allows us to keep snapshots in place. Thanks!

kentcdodds:
Try this one: https://github.com/joaogranado/jest-serializer-ansi

gnapse:
Done in #17, but somehow build is still failing over there, even when they all pass locally :(

kentcdodds:
Ugh.

gnapse:
There doesn't seem to be a definite solution for this problem in that travis issue discussion :(

Given that #17 remains failing for some other reason even after serializing snapshots without the ansi stuff, I'm leaning towards merging this PR, because it nevertheless fixes the issue reported in kentcdodds/dom-testing-library#50.

I'll try to dig a bit more and wait until later today, to see if a this can be solved. Otherwise I'll merge because we can always bring back the snapshots with proper serialization and/or travis settings fixed or whatever, in a later PR.

PR #27, opened 2018-07-05
fix: Types incorrectly pulling in .src

smacpherson64:
**What**:
A fix for types where it was incorrectly pulling in .src

**Why**:
It will cause errors in Typescript when trying to compile a project with jest-dom

**How**:
Removed the unnecessary reference

**Checklist**:
* [x] Documentation - N/A
* [x] Tests - N/A
* [x] Ready to be merged 

### DIFF
--- a/extend-expect.d.ts
+++ b/extend-expect.d.ts
@@ -1,5 +1,3 @@
-import {toContainElement} from './src'
-
 declare namespace jest {
   interface Matchers<R> {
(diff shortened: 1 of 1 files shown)


smacpherson64:
@kentcdodds Sorry, I accidentally added an invalid reference to the `extend-expect.d.ts` file which causes Typescript to not be able to compile.

smacpherson64:
Thank you! Again sorry for not catching it earlier!

PR #30, opened 2018-07-08
TypeScript: Convert matchers to instance methods

jgoz:
**What**:

- Convert matcher function definitions to instance methods (rather than function properties)
- Allow `RegExp` in `toHaveTextContent` typings 

**Why**:

- Jest's typings (on DefinitelyTyped) use instance methods for matchers &mdash; this difference is mostly cosmetic, but TypeScript does treat methods slightly differently from function properties with respect to generics when `strictFunctionTypes` is enabled
- `toHaveTextContent` implementation allows a RegExp to be passed instead of a string

Typings before:
Typings - Before

Typings after:
Typings - After

**How**:

- Updated extend-expect.d.ts

**Checklist**:

* [ ] Documentation N/A
* [ ] Tests N/A
* [x] Ready to be merged 

### DIFF
--- a/extend-expect.d.ts
+++ b/extend-expect.d.ts
@@ -7,6 +7,6 @@ declare namespace jest {
     toHaveAttribute(attr: string, value?: string): R
     toHaveClass(className: string): R
     toHaveStyle(css: string): R
-    toHaveTextContent(text: string): R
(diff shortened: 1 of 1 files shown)


gnapse:
This is great. Thanks!

PR #29, opened 2018-07-05
Allow SVGElements in matchers

jgoz:
Fixes #28 

**What**: Allows matchers to be used on `SVGElement` objects

**Why**: These matchers may be validly used on SVG elements

**How**: Added `instanceof SVGElement` check to `checkHtmlElement`

**Checklist**:

* [ ] Documentation N/A
* [x] Tests
* [x] Ready to be merged 
* [x] Added myself to contributors table

PR #36, opened 2018-07-11
docs: Updated README to Have Types

### LINKED ISSUE #33: Document supported types of arguments for all matchers in the README
Following up [this comment](https://github.com/gnapse/jest-dom/pull/24#issuecomment-403614587) and the discussion above it, I think it's important that we document explicitly in the README the supported types of arguments for all custom matchers provided by this library.

smacpherson64:
**What**:

Updating readme to have types

**Why**:
To complete #33 

**How**:
Added types (typescript types without return values)

**Checklist**:

* [x] Documentation N/A
* [x] Tests  N/A
* [x] Ready to be merged
* [x] Added myself to contributors table N/A

Closes #33

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,7 +205,7 @@ expect(getByTestId(container, 'count-value')).not.toHaveTextContent('21')
 toHaveAttribute(attr: string, value?: string)
 ```
 
-This allows you to check wether the given element has an attribute or not. You
+This allows you to check whether the given element has an attribute or not. You
 can also optionally check that the attribute has a specific expected value.
 
 ```javascript
@@ -231,7 +231,7 @@ expect(getByTestId(container, 'ok-button')).not.toHaveAttribute(
 toHaveClass(...classNames: string[])
 ```
 
-This allows you to check wether the given element has certain classes within its
+This allows you to check whether the given element has certain classes within its
 `class` attribute.
 
 ```javascript
--- a/README.md
+++ b/README.md
@@ -99,6 +99,10 @@ expect.extend({toBeInTheDOM, toHaveClass})
 
 ### `toBeInTheDOM`
 
+```
+toBeInTheDOM(container?: HTMLElement)
+```
+
 This allows you to assert whether an element present in the DOM container or not. If no DOM container is specified it will use the default DOM context.
 
 #### Using the default DOM container
@@ -137,6 +141,10 @@ expect(queryByTestId(container, 'ancestor')).not.toBeInTheDOM(
 
 ### `toBeEmpty`
 
+```
+toBeEmpty()
+```
+
 This allows you to assert whether an element has content or not.
 
 ```javascript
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -99,6 +99,10 @@ expect.extend({toBeInTheDOM, toHaveClass})
 
 ### `toBeInTheDOM`
 
+```
+toBeInTheDOM(container?: HTMLElement)
```
gnapse: Since #29 we also support `SGVElement`, which for some reason is a class of its own. I guess we missed reflecting this change in the typescript types too, so maybe we could take the opportunity to reflect the change in that file as well as in the README.

@jgoz since you worked in #29 and also have worked on TypeScript stuff in this repo, can you confirm the typescript signatures need to be updated to reflect that they support sag elements as well?

README.md
```diff
@@ -152,6 +160,10 @@ expect(queryByTestId(container, 'not-empty')).not.toBeEmpty()
 
 ### `toContainElement`
 
+```
+toContainElement(element: HTMLElement)
```
gnapse: Same thing about the `SVGElement` support mentioned above.

gnapse:
BTW @smacpherson64 I'm not sure this qualifies as a `fix:` commit. Maybe if we end-up needing to update the typings, then yes. But if it's only a README change, it should be `docs:`. This is for the squash-merge commit. There's no problem with these intermediate commits to have the `fix:` prefix. I think.

smacpherson64:
About the `docs:` sounds good, I will make sure to do that going forward!

gnapse:
If we need to update the typescript definitions, as I suspect, then it's a fix.

smacpherson64:
I think we should keep them separate and make a new issue to validate and fix the types. I am fairly sure we will need to add SVGElement as well:

```
document.createElementNS("http://www.w3.org/2000/svg", "svg") instanceof HTMLElement; // false

/* From: https://stackoverflow.com/questions/24174975/document-createelementsvg-instanceof-svgelement-is-false */
```

smacpherson64:
I added typescript to the code blocks as I do not know how adding them to the headers will affect linking.

README.md
```diff
@@ -207,6 +227,10 @@ expect(getByTestId(container, 'ok-button')).not.toHaveAttribute(
 
 ### `toHaveClass`
 
+```typescript
+toHaveClass(...classNames: string[])
+```
+
 This allows you to check wether the given element has certain classes within its
```
jgoz: Nitpick: `wether` should be `whether` (it's misspelled in a few spots)

README.md
jgoz: @gnapse That's a good question. I suppose it's possible that someone would be using an SVG element as a container, so maybe we should update the signature as follows (both in the readme and in definitions):

```ts
toBeInTheDOM(container?: HTMLElement | SVGElement)
```

gnapse: Good to know. Are you up for it @jgoz? If not don't worry, maybe @smacpherson64 or I can do it later.

jgoz: For sure, I can add it to my list today.

gnapse:
Yes, good point regarding the links. They'd be updated, since the TOC is auto-generated, and it would all work, but it's ok this way anyway. Feel free to merge this, and remember to set the `docs:` prefix. Thanks!

README.md
gnapse: Nice catch. Thanks!

PR #39, opened 2018-07-11
Change toHaveClass behaviour for empty class list

jgoz:
**What**:

- Require at least one expected class in `toHaveClass` (per https://github.com/gnapse/jest-dom/pull/31#issuecomment-404225469)
- `.not.toHaveClass()` asserts that the actual class list is empty (per https://github.com/gnapse/jest-dom/pull/31#issuecomment-404307031)

**Why**:

- Invoking `toHaveClass` without expected classes will always pass, but is probably indicative of a developer mistake.
- Asserting that a class list is empty (`.not.toHaveClass()`) was not previously possible using this library

**How**:

- Conditional result based on number of expected classes
  - `expected.length > 0` &rarr; previous behaviour
  - `expected.length === 0` and `isNot` &rarr; assert `received.length === 0`
  - `expected.length === 0` and `!isNot` &rarr; fail (at least one class required)

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged 
* [ ] Added myself to contributors table  N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,6 +249,17 @@ expect(getByTestId(container, 'delete-button')).not.toHaveClass('btn-link')
 // ...
 ```
 
+You must provide at least one class, unless you are asserting that an element
+does not have any classes.
+
+```javascript
+// ...
+// <button data-testid="no-classes">
+//   Delete item
+// </button>
+expect(getByTestId(container, 'no-classes')).not.toHaveClass()
+```
+
 ### `toHaveStyle`
 
 ```typescript
--- a/src/__tests__/to-have-class.js
+++ b/src/__tests__/to-have-class.js
@@ -14,6 +14,7 @@ test('.toHaveClass', () => {
       <svg data-testid="svg-spinner" class="spinner clockwise">
         <path />
       </svg>
+      <div data-testid="no-classes"></div>
     </div>
   `)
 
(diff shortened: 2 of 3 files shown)


gnapse:
`.not.toHaveClass()` seems the same as `.not.toHaveAttribute('class')` except maybe for a case like `<div class="">...</div>`, but I see no harm in having it. Good job!

gnapse:
Should we consider this a breaking change?

In case the answer is positive, we have a couple (or even maybe 3) upcoming breaking changes. I was thinking if we should bundle them in a `release` branch 🤔.

kentcdodds:
If you're planning on doing a breaking release anyway then this should probably go with that. I doubt anyone's had time to make this mistake and rely on it as a feature, but better to be safe :+1:

jgoz:
@gnapse technically it is a change in behaviour for `.toHaveClass()` but I don’t think there were any valid use cases for that to begin with. In other words, if this was published as a patch/minor, it might lead to test failures, but only for invalid tests that need rewriting.

kentcdodds:
That's fair @jgoz. I'd be fine either way.

jgoz:
BTW, this behaviour was possible before my params change - if you didn’t pass a class name, the string would be undefined and result in `[]` for `subset`, same as now.

jgoz:
I’m also fine either way 🙂

If there are other breaking changes queued up, then it would be fine to bundle this with them.

gnapse:
> BTW, this behaviour was possible before my params change - if you didn’t pass a class name, the string would be undefined and result in [] for subset, same as now.

I think this settles it as not a breaking change. A fix it is then.

Issue #34, opened 2018-07-10
Deprecate toBeInTheDOM in favor of toContainElement


gnapse:
### Describe the feature you'd like:

It's been previously noted that `toBeInTheDOM` does not really perform the check that its name suggests (kentcdodds/dom-testing-library#9, #3). This was partially addressed in #25, where also `toContainElement` was introduced.

However, `toBeInTheDOM` continues to feel not right, at least for me. It now has two different behaviors: if it receives an element as argument, it'll check that the `expect` argument is contained within that extra element. If it does not receive an extra element as argument, it behaves as a glorified `.toBeInstanceOf(HTMLElement)`, merely checking that the received `expect` argument is a DOM element, regardless of whether is contained anywhere.

### Suggested implementation:

I propose that we remove `toBeInTheDOM`, maybe with a deprecation period.

### Describe alternatives you've considered:

At the very least, we should update the documentation in the README about it. This current intro to this matcher is not accurate about what it does:

> This allows you to assert whether an element present in the DOM container or not. If no DOM container is specified it will use the default DOM context.

There's no default DOM context (whatever that means). This matcher's only task when no DOM container is specified, is to assert that the given element is indeed an HTML (or SVG since recently) element.

### Teachability, Documentation, Adoption, Migration Strategy:

We would need to update the documentation and provide users with alternatives moving forward:

One recommendation is to replace current uses with `toContainElement`.

Also, if users just want to check if an element exist, they can assert that what they have is not null or undefined. For instance, using `dom-testing-library` queries and regular jest matchers:

```js
expect(queryByTestId('ok-button')).toBeTruthy();
expect(queryByTestId('ok-button')).toBeInstanceOf(HTMLElement);
```

Also, this would be a breaking change, so a major version bump is involved and users would have to explicitly update to get the change and be really forced to replace its uses.

kentcdodds:
I'm in favor 👍

smacpherson64:
Same! Honestly toBeInTheDOM reads well but is only helpful if, like you said, it has a document context like `document.documentElement.contains`. Without it, it is kinda dangerous. 

E.G.: expect(document.createElement(‘div’)).not.toBeInTheDOM())

gnapse:
Good point. If we can make `expect(element).toBeInTheDOM()` to be equivalent to

```js
expect(document.documentElement).toContainElement(element)
```

[...]

PR #40, opened 2018-07-14
feat: Depreciate toBeInTheDOM with replacement

smacpherson64:
**What**:
Deprecating toBeInTheDOM
Adding toBeInTheDocument

**Why**:
Aiming to implement desires from conversation in issue #34

**How**:
Added a deprecate function to warn of deprecation
Added jsdoc to types to label as deprecated
Added toBeInTheDocument function
Updated documentation to have deprecated notices
Added toBeInTheDocument to documentation

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged

@gnapse, @jgoz, and @kentcdodds Here is the first attempt to do the changes we discussed in #34, when you have time please review. :-) !

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ expect(queryByTestId(container, 'not-empty')).not.toBeEmpty()
 toBeInTheDocument()
 ```
 
-This allows you to assert whether an element present in the document or not.
+This allows you to assert whether an element is present in the document or not.
 
 ```javascript
 // add the custom expect matchers once
--- a/README.md
+++ b/README.md
@@ -137,7 +137,8 @@ import 'jest-dom/extend-expect'
 // const svgElement = document.querySelector('[data-testid="html-element"]')
 // const detachedElement = document.createElement('div')
 
-expect(queryByTestId(container, 'count-value').toBeInTheDocument()
+expect(htmlElement).toBeInTheDocument()
+expect(svgElement).toBeInTheDocument()
 expect(detacthedElement).not.toBeInTheDocument()
 // ...
 ```
--- a/README.md
+++ b/README.md
@@ -131,9 +131,14 @@ This allows you to assert whether an element is present in the document or not.
 import 'jest-dom/extend-expect'
 
 // ...
-// <span data-testid="count-value">2</span>
+// document.body.innerHTML = `<span data-testid="html-element"><span>Html Element</span></span><svg data-testid="svg-element"></svg>`
+
+// const htmlElement = document.querySelector('[data-testid="html-element"]')
+// const svgElement = document.querySelector('[data-testid="html-element"]')
+// const detachedElement = document.createElement('div')
+
 expect(queryByTestId(container, 'count-value').toBeInTheDocument()
-expect(queryByTestId(container, 'count-value1')).not.toBeInTheDocument()
+expect(detacthedElement).not.toBeInTheDocument()
 // ...
 ```
 
@@ -363,9 +368,11 @@ expect(getByText(container, 'LINK')).not.toBeDisabled()
 
 ### `toBeInTheDOM`
 
-> Please use [`toBeInTheDocument`](#tobeinthedocument) for searching the entire document.
+> Note: The differences between `toBeInTheDOM` and `toBeInTheDocument` are significant. Replacing all uses of `toBeInTheDOM` with `toBeInTheDocument` will likely cause unintended consequences in your tests. Please make sure when replacing `toBeInTheDOM` to read through the replacements below to see which use case works better for your needs.
 
 > Please use [`toContainElement`](#tocontainelement) for searching a specific container.
+
+> Please use [`toBeInTheDocument`](#tobeinthedocument) for searching the entire document.
 
 ## Inspiration
 
--- a/README.md
+++ b/README.md
@@ -45,8 +45,8 @@ to maintain.
 - [Installation](#installation)
 - [Usage](#usage)
 - [Custom matchers](#custom-matchers)
-  - [`toBeInTheDOM`](#tobeinthedom)
   - [`toBeEmpty`](#tobeempty)
+  - [`toBeInTheDocument`](#tobeinthedocument)
   - [`toContainElement`](#tocontainelement)
   - [`toHaveTextContent`](#tohavetextcontent)
   - [`toHaveAttribute`](#tohaveattribute)
@@ -55,6 +55,8 @@ to maintain.
   - [`toHaveFocus`](#tohavefocus)
   - [`toBeVisible`](#tobevisible)
   - [`toBeDisabled`](#tobedisabled)
+- [Deprecated matchers](#deprecated-matchers)
+  - [`toBeInTheDOM`](#tobeinthedom)
 - [Inspiration](#inspiration)
 - [Other Solutions](#other-solutions)
 - [Guiding Principles](#guiding-principles)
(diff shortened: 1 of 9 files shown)


src/utils.js
```diff
@@ -66,4 +66,29 @@ function matches(textToMatch, node, matcher) {
...
-export {checkHtmlElement, getMessage, matches}
+function deprecate(name, replacementText) {
+  const message = getDepreciationMessage(name, replacementText)
+
+  // Notify user that they are using deprecated functionality.
+  // eslint-disable-next-line no-console
+  console.warn(message)
+}
+
+function getDepreciationMessage(name, replacementText) {
```
jgoz: Does this need to be exposed as a separate function at all? I know it's used in the tests, but I would rather see the full message typed out in the tests than use this function in both spots. Otherwise we never really test the output of this function.

My preference would be to remove this function and inline the message creation in `deprecate`:


```js
function deprecate(name, replacementText) {
  // Notify user that they are using deprecated functionality.
  // eslint-disable-next-line no-console
  console.warn(
    `Warning: ${name} has been deprecated and will be removed in future updates.`,
     replacementText
  )
}

File: src/__tests__/utils.js [hunk 2]
```diff
@@ -0,0 +1,15 @@
+import {deprecate, getDepreciationMessage} from '../utils'
+
+test('deprecate', () => {
+  const spy = jest.spyOn(console, 'warn').mockImplementation(() => {})
+
+  const messageWithReplacement = getDepreciationMessage('test', 'test')
```
jgoz: As I commented below, I think we should type out the full expected message here, rather than rebuilding it again with the function.

src/to-be-in-the-document.js
```diff
@@ -0,0 +1,24 @@
+import {matcherHint, printReceived} from 'jest-matcher-utils'
+import {checkHtmlElement} from './utils'
+
+export function toBeInTheDocument(element) {
+  checkHtmlElement(element, toBeInTheDocument, this)
+
+  return {
+    pass: document.documentElement.contains(element),
```
jgoz: Should we check whether `document` and `document.documentElement` exist? I know `jsdom` will have them, but if I accidentally used this matcher in the `node` environment, it would be nice to get a descriptive error message rather than "cannot read property 'documentElement' of undefined".

smacpherson64:
@jgoz Thanks for the great feedback!

src/utils.js
smacpherson64: Sounds good, I will aim to remove this and simplify the check to just make sure console.warn was called.

smacpherson64: I will remove it, but will also first try to match the consoles, which shouldn't be too crazy.

src/to-be-in-the-document.js
smacpherson64: Hrm, sounds good, agreed on developer experience. It would be nice to get a clear error. Does the following sound good @jgoz?

`document is undefined on global but is required for the toBeInTheDocument matcher`.

`documentElement is undefined on document but is required for the toBeInTheDocument matcher.`

jgoz: 👍  the first message should be good enough for both cases.

README.md
```diff
@@ -87,77 +89,56 @@ Alternatively, you can selectively import only the matchers you intend to use,
...
-### `toBeEmpty`
+### `toBeInTheDocument`
 
 ```typescript
-toBeEmpty()
+toBeInTheDocument()
 ```
 
-This allows you to assert whether an element has content or not.
+This allows you to assert whether an element present in the document or not.
```
jgoz: Missing `is` before `present`

smacpherson64: Nice find! Kk, fixed. (I hope you have an awesome vacation!)

smacpherson64:
@jgoz: Progress on better error message on document for developer experience:

## TDLR
I am not sure that jest-dom can be used without the jsdom environment. I get an error when trying to test without document:

```
... jest-dom\node_modules\jsdom\lib\jsdom\browser\Window.js:598
      window.document.addEventListener("load", () => {
```

I believe it is pulled in here:
https://github.com/gnapse/jest-dom/blob/master/jest.config.js#L4

## Details
To accomplish get a better error message when the document is missing, I created a `checkDocumentKey` utility similar to checkHtmlElement:

```javascript
class InvalidDocumentError extends Error {
  constructor(message, matcherFn) {
    super()

    /* istanbul ignore next */
    if (Error.captureStackTrace) {
      Error.captureStackTrace(this, matcherFn)
    }

    this.message = message
  }
}

function checkDocumentKey(key, matcherFn) {
  if (typeof document === 'undefined') {
    throw new InvalidDocumentError(
      `document is undefined on global but is required to use ${
        matcherFn.name
      }.`,
      matcherFn,
    )
  }

  if (typeof document[key] === 'undefined') {
    throw new InvalidDocumentError(
      `${key} is undefined on document but is required to use ${
        matcherFn.name
      }.`,
      matcherFn,
    )
  }
}
```
I can successfully get all the tests to pass however, getting coverage up to 100% requires testing without the document. When I remove the document from global, the util test throws an error.

### Tests that Cause Error
```javascript
describe('checkDocumentKey without document', () => {
  beforeEach(() => {
    global.previousDocument = {...document}
    delete global.document
  })

  test('checkDocumentkey', () => {
    expect(() =>
      checkDocumentKey('documentElement', matcherMock),
    ).toThrowError()
  })

  afterEach(() => {
    global.document = global.previousDocument
    delete global.previousDocument
  })
})
```

### Error
```
... jest-dom\node_modules\jsdom\lib\jsdom\browser\Window.js:598
      window.document.addEventListener("load", () => {
```

smacpherson64:
Well, I was wrong haha, there is always another way. @jgoz I have updated the concept and it should now check for undefined document or a specific key on document.

README.md
```diff
@@ -87,77 +89,56 @@ Alternatively, you can selectively import only the matchers you intend to use,
...
 // add the custom expect matchers once
 import 'jest-dom/extend-expect'
 
 // ...
-// <span data-testid="not-empty"><span data-testid="empty"></span></span>
-expect(queryByTestId(container, 'empty')).toBeEmpty()
-expect(queryByTestId(container, 'not-empty')).not.toBeEmpty()
+// <span data-testid="count-value">2</span>
+expect(queryByTestId(container, 'count-value').toBeInTheDocument()
+expect(queryByTestId(container, 'count-value1')).not.toBeInTheDocument()
```
gnapse: I think that `expect(nullOrUndefined).toBeInTheDocument()` would fail before even having the chance to actually check for presence in the document. So this second example is not accurate because `queryByTestId(container, 'count-value1')` will return `null` or `undefined`, and neither of those is a html element.

I think that `expect(something).not.toBeInTheDocument()` only passes if `something` is an html or sag element, but it's not attached to the document. It won't work at all for things that are not html or svg elements.

So in this sense, it's not a direct replacement for `toBeInTheDOM`, and I think that's a good thing. The migration path is not simply "replace all uses of `toBeInTheDOM` with `toBeInTheDocument`.

Not sure though on how to approach this in the README example to reflect on both cases. And also I think we need a test that asserts that `toBeInTheDocument` indeed fails when given `null` or `undefined`.

smacpherson64: Sounds Good! I will work to take care of these!

Thank you @gnapse!

smacpherson64:
Hey @gnapse!

I believe the following should be complete:

- Updated the README to have better examples for toBeInTheDocument (using examples from the tests).
- Added a disclaimer aimed to emphasize that `toBeInTheDocument` is not a direct replacement of `toBeInTheDOM`.
- Added null and undefined tests

Please let me know of any other change you desire!

README.md
```diff
@@ -131,9 +131,14 @@ This allows you to assert whether an element is present in the document or not.
...
 
 // ...
-// <span data-testid="count-value">2</span>
+// document.body.innerHTML = `<span data-testid="html-element"><span>Html Element</span></span><svg data-testid="svg-element"></svg>`
+
+// const htmlElement = document.querySelector('[data-testid="html-element"]')
+// const svgElement = document.querySelector('[data-testid="html-element"]')
+// const detachedElement = document.createElement('div')
+
 expect(queryByTestId(container, 'count-value').toBeInTheDocument()
```
gnapse: I would've thought this line was also going to be replaced to use some of the other elements you define above that are indeed in the document.

I'd even insist in changing it merely on the ground that this is referencing a helper function in dom-testing-library, probably since this README was created based on that other library's README.

smacpherson64: I am sorry! Great catch!

smacpherson64:
@gnapse again, Great catch!

Do the updates match what you were desiring?

gnapse:
Hmmm, merging this PR did not create a new release in npm, even though in [the releases page](https://github.com/gnapse/jest-dom/releases) it seems like something did happen. The latest two releases appear but just as tags, not as full features releases. And indeed, in npm the two new releases do not appear. It's stuck at v1.8.1.

I sense this has something to do with the recent tokens invalidations in npmjs.com, but I'm not sure how to update semantic-release's token. Care to chime in @kentcdodds?

kentcdodds:
Oh yes, I manually generated a new token on npm and updated the NPM_TOKEN variable on all my Travis jobs. Took a little bit of time (I have quite a few packages) but wasn't too much of a challenge. Let me know if you need help.

PR #42, opened 2018-07-24
feat: assert whether a string Element is in the document OR contained…

6ewis:
**What**:

 Assert whether a string representing a  HTML Element is in the document: `expect("<div>Example</div>").toBeInTheDocument()`

Assert whether a string representing a HTML element is contained in another Element: `expect(container).toContainElement(<div> Example </div>)`

**Why**:

It's more intuitive to expect the methods `toContainElement()` and `toBeInTheDocument()` to expect HTML wrapped in strings.  Also, there are use cases where it makes more sense. Picture a component that receives a dynamic HTML content as an input and parse it via a HTML parser  - we could then test that the newly added DOM Element are in the document. It is more declarative - we are not only limited to query the DOM to check if an element is in the DOM. 

**How**:

Simply By checking the type of the argument, if it is a string then we verify that it is a valid HTML element first then we check that it is included in the html DOM.

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged 

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@
 [![downloads][downloads-badge]][npmtrends]
 [![MIT License][license-badge]][license]
 
-[![All Contributors](https://img.shields.io/badge/all_contributors-17-orange.svg?style=flat-square)](#contributors)
+[![All Contributors](https://img.shields.io/badge/all_contributors-18-orange.svg?style=flat-square)](#contributors)
 [![PRs Welcome][prs-badge]][prs]
 [![Code of Conduct][coc-badge]][coc]
 
@@ -48,6 +48,7 @@ to maintain.
   - [`toBeEmpty`](#tobeempty)
   - [`toBeInTheDocument`](#tobeinthedocument)
   - [`toContainElement`](#tocontainelement)
+  - [`toContainHtml`](#tocontainhtml)
   - [`toHaveTextContent`](#tohavetextcontent)
   - [`toHaveAttribute`](#tohaveattribute)
   - [`toHaveClass`](#tohaveclass)
@@ -166,6 +167,27 @@ expect(ancestor).toContainElement(descendant)
 expect(descendant).not.toContainElement(ancestor)
 // ...
 ```
+
+### `toContainHtml`
+
+```typescript
+toContainHtml(htmlText: String)
+```
+
+Assert whether a string representing a HTML element is contained in another Element:
+
+```javascript
+// add the custom expect matchers once
+import 'jest-dom/extend-expect'
+
+// ...
+// <span data-testid="parent"><span data-testid="child"></span></span>
+const parent = queryByTestId('parent')
+expect(parentElement).toContainHtml('<span data-testid="child"></span>')
+// ...
+```
+
+> Note: Chances are you probably do not need to use this method. We encourage testing from the perspective of how the user perceives the app in a browser. That's why testing against a specific DOM structure is not advised. It might come in handy in rare cases such as if your users manipulate the DOM such as a content management tool - where users are expected to type custom HTML. please use: [`toContainElement`](#tocontainelement)
 
 ### `toHaveTextContent`
 
@@ -401,7 +423,7 @@ value is indeed an `HTMLElement` you can always use some of
 
 ```js
 expect(document.querySelector('.ok-button')).toBeInstanceOf(HTMLElement)
-expect(document.querySelector('.cancel-button')).toBeThruthy();
+expect(document.querySelector('.cancel-button')).toBeThruthy()
 ```
 
 > Note: The differences between `toBeInTheDOM` and `toBeInTheDocument` are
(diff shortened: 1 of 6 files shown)


gnapse:
This looks interesting. I wonder though if this would be resilient to subtle changes in the string representation of the element expected to be found. Also, how resilient is this to changes of whitespace in the string representation of the element to be found.

For instance, consider the following example:

```js
// <div><h1>Hello</h1> ...more stuff</div>
expect('<h1>Hello</h1>').toBeInTheDocument();
```

Would the above test still pass after some code is introduced that attaches, for example, a `data-` attribute to that element? Or does the user need to provide an exact html representation of the element, with all its attributes?

```js
// <div><h1 data-outline="1.1">Hello</h1> ...more stuff</div>
expect('<h1>Hello</h1>').toBeInTheDocument();
```

The other caveat I see could happen is with whitespace:

```js
// <div><h1>  
//      Hello
//
// World
//     </h1> ...more stuff</div>
expect('<h1>Hello World</h1>').toBeInTheDocument();
```

Not saying it is wrong if any of the two examples above do not pass. I just wanted to bring it up for discussion, and if that's the case, and if this is to be merged, we need to document those caveats, or better yet, see if we can mitigate some of their effects.

kentcdodds:
I worry that this would make the tests to tied to the DOM implemention. What if I put some text in a span? Doesn't change the way it looks, but it does break the test. I personally would not use it recommend this feature.

jgoz:
I understand the motivation for this change, but I think its limitations make it too brittle, as @gnapse and @kentcdodds pointed out above. Maybe if the behaviour was closer to [enzyme's `containsMatchingElement`](http://airbnb.io/enzyme/docs/api/ShallowWrapper/containsMatchingElement.html) and compared attribute subsets, was insensitive to whitespace, etc., it would be more broadly useful. But that is a lot more effort.

6ewis:
@gnapse I think you made fair points, but isn't it it the same concerns with getByText and white spaces? 
@kentcdodds I don't find it to me more tied to the DOM implementation than the current implementation. Can you expand on your thoughts ? I'm not sure I follow. I came up with this after writing expect(container.outerHTML).toMatch() a dozen times in my use case, I wanted to make sure that whatever HTML my component receives is indeed in the DOM
@jgoz I don't believe that there's anything wrong by being inherently 'exact'

jgoz:
> @jgoz I don't believe that there's anything wrong by being inherently 'exact'

That's true, and there are other parts of this library that I wish were more exact by default (such as `toHaveTextContent`). My issue isn't directly with exactness, it's with the effects thereof &mdash; comparing the exact DOM representation of an element can lead to a lot of churn in your assertions as things evolve because changes that might be unrelated to particular tests are more likely to cause those tests to fail.

Another consequence of exactness, in this case, is that the tests are verifying details about the structure of your markup that should be irrelevant to the user. I.e., as a user, I don't care if the "user name" element uses a `div` or a `span`, I just care that it contains my username. You as the developer may care about the structure, but assertions like that go against the principles of the testing libraries that spawned `jest-dom` (`react-testing-library`, `dom-testing-library`).

6ewis:
@jgoz let's agree to disagree. In my use case my user directly manipulates html from a platform called contentful. Not all users are created equal so quite the opposite it seems that I'm testing what they care about. They expect their html to be inserted in the DOM. I believe context matters, and generalizing that exactness is always testing implementation details is not a statement I agree with.

gnapse:
@6ewis your use case sounds like you have components that receive html as a string (provided by a user in the real use of those components in your app) and they render that html within them, and you want to assert that they did so and used that html verbatim. Is that so? Sounds interesting!

It kinda puts our concerns to test, I have to admit. What we've all raised here is not to be easily dismissed though. `jest-dom` was spawned off `react-testing-library`, and the guiding principle behind it is to encourage testing from the perspective of how the user perceives the app in a browser. That's why testing against a specific DOM structure is usually considered counterproductive. For example, the user normally cares to see a label saying "Your message was sent successfully!". They do not care if that message internally is a `<div>Your message...</div>` or if it's enclosed in a `<span>`. That's what we mean with our concerns, and that's why your feature request seems to go against those guiding principles. It may so that your use case is different, and we can discuss it. In your use case it may be valid, but we're worried that this will be misused for use cases like the successful message notification above.

smacpherson64:
This is a really good conversation! The concept of how contentful is being used sounds awesome and is a fun use case! At first glance, it looks like a perfect opportunity for snapshot testing. 

If the PR is to merged, I prefer we keep the current interfaces, for consistency with the rest of the matchers, and don't overload the selectors to accept strings. Instead, I think a separate selector would be clearer: E.G.: 
```
expect(container: HTMLElement).toContainHTML(htmlText: string)
```

gnapse:
Hmmm, good point about snapshot testing. Wouldn't that work for you @6ewis?

Otherwise, the `toContainHTML` proposal is a very good idea. This would allow us to better document the caveats and warn about misuse, without affecting with such warnings the current less controversial use of the matchers being modified here.

6ewis:
@gnapse I shy away from snapshot testing in general because people tend to update large snapshots without paying much attention; I prefer more explicit assertions. More over, we cannot control what html the user will put - it's up to the user to decide whether her html is valid or not. We only sanitize the html. 

Thank you guys. I'll leave it to you to decide whether it's worth it to merge it or not. I like @smacpherson64 suggestions so I updated my PR and the documentation. However, it could also be good to see if it's a requested feature in the coming months. It may be too much of a rare case. The guiding principle of react testing lib took me here, so we should never lose sight of it.

README.md
```diff
@@ -167,6 +168,27 @@ expect(descendant).not.toContainElement(ancestor)
...
+
+// ...
+// <span data-testid="parent"><span data-testid="child"></span></span>
+const parent = queryByTestId('parent')
+expect(parentElement).toContainHtml('<span data-testid="child"></span>')
+// ...
+```
+
+> Note: Chances are you probably do not need to use this method. We encourage testing from the perspective of how the user perceives the app in a browser. That's why testing against a specific DOM structure is not advised. It might come in handy in rare cases such as if your users manipulate the DOM such as a content management tool - where users are expected to type custom HTML. please use: [`toContainElement`](#tocontainelement)
+
```
gnapse: I came up with an alternative that perhaps got too lengthy, so feel free to just take ideas from it. Cause I still think the text above is missing some key features:

> Chances are you probably do not need to use this matcher. We encourage testing from the perspective of how the user perceives the app in a browser. That's why testing against a specific DOM structure is not advised.
>
> It could be useful in situations where the code being tested renders html that was obtained from an external source, and you want to validate that that html code was used as intended.
>
> It should not be used to check DOM structure that you control. **A good rule of thumb would be to not use it if you find yourself passing a string literal as the expected html.** In such cases please obtain actual references to the elements you want to check for (using `document.querySelector` or other equivalent methods) and use [`toContainElement`](#tocontainelement) instead.

README.md
```diff
@@ -167,6 +168,27 @@ expect(descendant).not.toContainElement(ancestor)
 // ...
 ```
 
+### `toContainHtml`
+
+```typescript
+toContainHtml(htmlText: String)
```
jgoz: `String` -> `string`

README.md
```diff
@@ -167,6 +168,27 @@ expect(descendant).not.toContainElement(ancestor)
...
+import 'jest-dom/extend-expect'
+
+// ...
+// <span data-testid="parent"><span data-testid="child"></span></span>
+const parent = queryByTestId('parent')
+expect(parentElement).toContainHtml('<span data-testid="child"></span>')
+// ...
+```
+
+> Note: Chances are you probably do not need to use this method. We encourage testing from the perspective of how the user perceives the app in a browser. That's why testing against a specific DOM structure is not advised. It might come in handy in rare cases such as if your users manipulate the DOM such as a content management tool - where users are expected to type custom HTML. please use: [`toContainElement`](#tocontainelement)
```
jgoz: You could probably leave out "It might come in handy in rare cases such as if your users manipulate the DOM such as a content management tool - where users are expected to type custom HTML".

And capital `P` on "please".

src/to-contain-html.js
```diff
@@ -0,0 +1,24 @@
+import {matcherHint, printReceived} from 'jest-matcher-utils'
+import {checkHtmlElement} from './utils'
+
+export function toContainHtml(container, htmlText) {
```
jgoz: This currently does not match the capitalization in the typescript definition. I suggest renaming everything to `toContainHTML`.

src/to-contain-html.js
```diff
@@ -0,0 +1,24 @@
...
+  const stringToHtml = item =>
```
jgoz: You could move this out of `toContainHTML` and combine it with `checkHtmlElement`:

```js
function checkHtmlText(htmlText, ...args) {
  const htmlElement = typeof htmlText === 'string'
    ? new DOMParser().parseFromString(htmlText, 'text/html').body.firstChild
    : null
  checkHtmlElement(htmlElement, ...args)
}
```

And then use it instead of `checkHtmlElement` below:

```js
checkHtmlElement(container, toContainHtml, this)
checkHtmlText(htmlText, toContainHtml, this)
```

src/__tests__/to-contain-html.js
```diff
@@ -0,0 +1,48 @@
+import {render} from './helpers/test-utils'
+
+/* eslint-disable max-statements */
```
jgoz: Is this still needed?

src/__tests__/to-contain-html.js
```diff
@@ -0,0 +1,48 @@
...
+  const parent = queryByTestId('parent')
+  const child = queryByTestId('child')
+  const nonExistantElement = queryByTestId('not-exists')
+  const fakeElement = {thisIsNot: 'an html element'}
+  const stringChildElement = '<span data-testid="child"></span>'
+  const svgElement = queryByTestId('svg-element')
+
+  expect(grandparent).toContainHtml(stringChildElement)
+  expect(parent).toContainHtml(stringChildElement)
+  expect(child).toContainHtml(stringChildElement)
```
jgoz: Can you add positive test cases for `.not`?

```js
const nonExistantString = '<div data-testid="child"></div>'
expect(grandparent).not.toContainHTML(nonExistantString)
expect(parent).not.toContainHTML(nonExistantString)
expect(child).not.toContainHTML(nonExistantString)
```

jgoz: And then add the corresponding negative test cases below?

```js
expect(() =>
  expect(grandparent).toContainHTML(nonExistantString),
).toThrowError()
// etc.
```

6ewis:
@jgoz @gnapse updated 👍

README.md
```diff
@@ -167,6 +168,31 @@ expect(descendant).not.toContainElement(ancestor)
...
+
+// ...
+// <span data-testid="parent"><span data-testid="child"></span></span>
+const parent = queryByTestId('parent')
+expect(parentElement).toContainHTML('<span data-testid="child"></span>')
+// ...
+```
+
+> Chances are you probably do not need to use this matcher. We encourage testing from the perspective of how the user perceives the app in a browser. That's why testing against a specific DOM structure is not advised.
+
```
gnapse: It's important to keep the `>` in front of these empty lines, so it all appears as a single note. The same applies to the next empty line below.

src/__tests__/to-contain-html.js
```diff
@@ -0,0 +1,64 @@
...
+    </span>
+    `)
+
+  const grandparent = queryByTestId('grandparent')
+  const parent = queryByTestId('parent')
+  const child = queryByTestId('child')
+  const nonExistantElement = queryByTestId('not-exists')
+  const fakeElement = {thisIsNot: 'an html element'}
+  const stringChildElement = '<span data-testid="child"></span>'
+  const nonExistantString = '<span> Does not exists </span>'
```
gnapse: (nit-picking alert!) should say "Does not exist" without the final "s".

Also, I still do not see any test about what happens if you pass invalid html, as in `<div>hello</p>`. I'm still not clear if in that case it fails because it won't find the invalid html in the container, or simply because it will fail to parse the invalid html in the first place.

6ewis:
@gnapse anything else?

README.md
```diff
@@ -188,9 +188,9 @@ expect(parentElement).toContainHTML('<span data-testid="child"></span>')
 ```
 
 > Chances are you probably do not need to use this matcher. We encourage testing from the perspective of how the user perceives the app in a browser. That's why testing against a specific DOM structure is not advised.
-
+> \
 > It could be useful in situations where the code being tested renders html that was obtained from an external source, and you want to validate that that html code was used as intended.
-
+> \
```
gnapse: I'm sorry for nit-picking, but why do you need the `\`? Without it you get separate praragraphs, which was my intention.

6ewis:
@gnapse done

Issue #46, opened 2018-07-28
toContainElement should accept null or undefined


gnapse:
I got to this idea when I realized that is not always possible to negate `.toContainElement`:

```js
expect(container).not.toContainElement(document.querySelector('.button'));
```

If indeed there's no element with `class="button"`, the above will still fail. But not because the matcher finds the assertion to not be true. It fails because `toContainElement` does not allow `null` or `undefined` as argument.

I wanted to propose changing this matcher so that it will accept `null` and `undefined`.

smacpherson64:
That is a good idea! What should the message say?

smacpherson64:
@gnapse is the following what you are desiring for?

```
expect(container).toContainElement(null) // false
expect(container).toContainElement(undefined) //false
expect(container).not.toContainElement(null) // true
expect(container).not.toContainElement(undefined) // true
```

gnapse:
Yes. I realize that it may look weird, and that's why I raised this as an issue first, to discuss it. But the thing is, it's weirder not to be able to negate a `.toContainElement`. Consider the following test around a component that loads data and you want to assert that it first renders a loading indicator, and then it removes it and renders data:

```js
// render stuff
expect(container).toContainElement(queryByText('Loading...'));
// wait for mocked request to resolve
expect(container).not.toContainElement(queryByText('Loading...'));
```

You cannot do the above because `toContainElement` will fail even before attempting to assert anything. You'd have to use a different matcher for the negative case. I think is perfectly fine for `toContainElement` not to pass when given null, and that it passes when it's invoked in the `.not` form with null.

jgoz:
@gnapse I agree. I have almost identical code to what you posted, except that the "not" case uses `toBeNull` now. It would be great to change it to your example, for clarity's sake.

gnapse:
Ok, seems there is a consensus. I wonder if this is a breaking change? Also, I can take it later this week, but if anyone else wants to give it a crack, go for it.

smacpherson64:
I can take a crack at this one! I think to be safe we should label it as a breaking change. Technically, the tests would perform the same, but the functionality is changing a bit.

jgoz:
If you go by the type definitions, it is not a breaking change since it would simply involve allowing `null` and `undefined` where they were previously not allowed. Also, any previously usages of `toContainElement(null)` would be throwing errors, so it's unlikely that they are being used.

jgoz:
[...]

PR #47, opened 2018-08-01
fix: Set toBeInTheDocument and toContainElement to be nullable

smacpherson64:
**What**:
- Set toBeInTheDocument and toContainElement to allow null and undefined.
- Adjusted response to make the failing text clearer

**Why**:
On issue #46, expressed the desire to be able to use null in toBeInTheDocument and toContainElement. This is useful when querying for elements that could result in null. It allows code like the following:

```js
// queryByText('Loading...') = null;
expect(queryByText('Loading...')).not.toBeInTheDocument()
expect(container).not.toContainElement(queryByText('Loading...'))
```

Instead of:

```js
expect(queryByText('Loading...').toBeNull();
```

**How**:
- Updated the elements in the matchers to be conditional
- Updated the matching text to make it clearer
- Updated the types to allow null and undefined

**Checklist**:
* [x] Documentation
* [x] Tests
* [x] Ready to be merged
* [x] Added myself to contributors table N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,7 @@ expect(detacthedElement).not.toBeInTheDocument()
 ### `toContainElement`
 
 ```typescript
-toContainElement(element: HTMLElement | SVGElement)
+toContainElement(element?: HTMLElement | SVGElement | null)
 ```
 
 This allows you to assert whether an element contains another element as a descendant or not.
--- a/README.md
+++ b/README.md
@@ -143,7 +143,8 @@ expect(htmlElement).toBeInTheDocument()
 expect(svgElement).toBeInTheDocument()
 expect(detacthedElement).not.toBeInTheDocument()
 expect(nonExistantElement).not.toBeInTheDocument()
-// ...```
+// ...
+```
 
 > Note: This will not find detached elements. The element must be added to the document to be found. If you desire to search in a detached element please use: [`toContainElement`](#tocontainelement)
 
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ expect(nonExistantElement).not.toBeInTheDocument()
 ### `toContainElement`
 
 ```typescript
-toContainElement(element?: HTMLElement | SVGElement | null)
+toContainElement(element: HTMLElement | SVGElement | null)
 ```
 
 This allows you to assert whether an element contains another element as a descendant or not.
--- a/README.md
+++ b/README.md
@@ -136,13 +136,14 @@ import 'jest-dom/extend-expect'
 
 // const htmlElement = document.querySelector('[data-testid="html-element"]')
 // const svgElement = document.querySelector('[data-testid="svg-element"]')
+// const nonExistantElement = document.querySelector('does-not-exist')
 // const detachedElement = document.createElement('div')
 
 expect(htmlElement).toBeInTheDocument()
 expect(svgElement).toBeInTheDocument()
 expect(detacthedElement).not.toBeInTheDocument()
-// ...
-```
+expect(nonExistantElement).not.toBeInTheDocument()
+// ...```
 
 > Note: This will not find detached elements. The element must be added to the document to be found. If you desire to search in a detached element please use: [`toContainElement`](#tocontainelement)
 
@@ -162,9 +163,11 @@ import 'jest-dom/extend-expect'
 // <span data-testid="ancestor"><span data-testid="descendant"></span></span>
 const ancestor = queryByTestId(container, 'ancestor')
 const descendant = queryByTestId(container, 'descendant')
+const nonExistantElement = queryByTestId(container, 'does-not-exist')
 
 expect(ancestor).toContainElement(descendant)
 expect(descendant).not.toContainElement(ancestor)
+expect(ancestor).not.toContainElement(nonExistantElement)
 // ...
 ```
 
--- a/extend-expect.d.ts
+++ b/extend-expect.d.ts
@@ -8,7 +8,7 @@ declare namespace jest {
     toBeVisible(): R
     toBeEmpty(): R
     toBeDisabled(): R
-    toContainElement(element?: HTMLElement | SVGElement | null): R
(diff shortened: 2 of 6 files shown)


src/to-contain-element.js
```diff
@@ -1,9 +1,16 @@
...
+  stringify,
+  RECEIVED_COLOR as receivedColor,
+} from 'jest-matcher-utils'
 import {checkHtmlElement} from './utils'
 
 export function toContainElement(container, element) {
   checkHtmlElement(container, toContainElement, this)
-  checkHtmlElement(element, toContainElement, this)
+
+  if (element) {
```
jgoz: Using a truthy check like this will mean these test cases will pass:

```js
expect(container).not.toContainElement(false);
expect(container).not.toContainElement(0);
expect(container).not.toContainElement("");
```

Is that the behaviour we want? (@gnapse)

jgoz:
Looks like the build is failing due to `localStorage is not available for opaque origins`. I just saw this in my projects and it's caused by a JSDOM upgrade. Setting `"testURL": "http://localhost/"` in the jest config fixes it.

src/to-contain-element.js
smacpherson64: Hrm, awesome point!

`element.contains` should throw a TypeError for those items (false, 0, and '') instead of returning false.

```
E.G.:
document.body.contains(0)
document.body.contains(false)
document.body.contains()
```

smacpherson64: @jgoz should we intentionally throw an error if it is not `null`, `undefined` or `HTMLElement`?

Something like:

`toBeInTheDocument expects an HTMLElement, null, or undefined but received 0`

smacpherson64: Kk, running through a test on jsdom, it appears it throws errors like chrome and FF:

```
test('if `element.contains` throw an error for falsey values', () => {
	expect(() => document.documentElement.contains(0)).toThrowError();
	expect(() => document.documentElement.contains('')).toThrowError();
	expect(() => document.documentElement.contains(false)).toThrowError();
});
```

jgoz: I was thinking it could fall through to `checkHtmlElement`:

```js
if (element !== null && element !== undefined) {
  checkHtmlElement(element, ...)
}
```

smacpherson64: That is a solid plan! I will aim to update tonight!

smacpherson64:
@jgoz Updated element validation logic to what we discussed.

gnapse:
If we add that example, it should be using `document.querySelector`. `queryByTestId` is a `dom-testing-library`-specific api, and I'd prefer the README to be more oriented to the general use case.

jgoz:
@gnapse Good point. The existing examples use `queryByTestId`, which is why I suggested it. But I agree with you.

gnapse:
Yes, I suspected that we may have some uses here and there. That's something I've been meaning to take care of. Not a big deal anyway.

smacpherson64:
Sounds good! My aim is to add the null example in the `readme.md` (using document.querySelector) by tonight. @jgoz and @gnapse, if you see any other changes you would like let me know. Thanks for letting me be part!

smacpherson64:
For the current time, I matched the tests to the surrounding code to make it consistent.

README.md
```diff
@@ -136,13 +136,14 @@ import 'jest-dom/extend-expect'
...
+// const nonExistantElement = document.querySelector('does-not-exist')
 // const detachedElement = document.createElement('div')
 
 expect(htmlElement).toBeInTheDocument()
 expect(svgElement).toBeInTheDocument()
 expect(detacthedElement).not.toBeInTheDocument()
-// ...
-```
+expect(nonExistantElement).not.toBeInTheDocument()
+// ...```
```
gnapse: The three back-ticks need to go on the following line

extend-expect.d.ts
```diff
@@ -8,7 +8,7 @@ declare namespace jest {
     toBeVisible(): R
     toBeEmpty(): R
     toBeDisabled(): R
-    toContainElement(element: HTMLElement | SVGElement): R
+    toContainElement(element?: HTMLElement | SVGElement | null): R
```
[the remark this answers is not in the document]
gnapse: And sorry for all the back and forth. Other than this minor detail (that's not decided yet, it's up for discussion) I think this is ready.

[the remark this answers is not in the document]
smacpherson64: Seriously, I really enjoy the feedback from you and @jgoz! The conversations usually make the end result better! I hope to tackle this by tonight.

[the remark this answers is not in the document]
jgoz: @gnapse good call!

smacpherson64:
@gnapse Just to make sure `fix:` sets it as a non-breaking change, where `feat:` is a breaking change?

I don't think we came to the conclusion:
https://github.com/gnapse/jest-dom/issues/46#issuecomment-409282459

I was thinking we should be safe and make it a breaking change, However, @jgoz made a great point:
> If you go by the type definitions, it is not a breaking change since it would simply involve allowing null and undefined where they were previously not allowed. Also, any previously usages of toContainElement(null) would be throwing errors, so it's unlikely that they are being used.

gnapse:
But `feat:` are not breaking changes. `feat:` just increases the minor version number, e.g. `1.11.1` would be followed by `1.12.0`.

A breaking change would make `1.11.1` be followed by `2.0.0`. And to do one, it's not enough with the prefix in the commit message. You'd have to include the words `BREAKING CHANGE` in the merge commit message. That's it not what I'm thinking.

I think this is a new feature, not a breaking change. A new feature is anything that introduces a new API, or adds a new capability to an existing API, without changing the current valid uses of any API. In this case I agree with @jgoz that this is not a breaking change, because these matchers didn't accept `null` as an argument, and now the do. So no one could've been using them in that way anyway.

smacpherson64:
I am good with whatever you think @gnapse on this!

Issue #52, opened 2018-08-17
Discussion - extending matchers to handle NodeList


smacpherson64:
### Describe the feature:
Would it make sense to extend the current matchers to handle NodeLists? Is there a need for this? I was thinking about `queryAllBy*` and `querySelectorAll` and wondering if it would be helpful to pass them directly to the matchers.

For example:
```javascript
expect(queryAllByText('text')).toBeVisible();
expect(document.querySelectorAll('button')).toBeDisabled();
```

If all elements match the matcher, then it passes. If any of the elements do not match then it fails.

kentcdodds:
I'm in favor

gnapse:
Sounds good. I'm in favor too.

I've got az question though: what would happen if the list of elements is empty?

The regular rules of logic would dictate that in that case an assertion passes. It gets even more interesting when you consider the negative case too, which should also pass, according to a strict logic:

```js
// allButtons is an empty list in a page without buttons
const allButtons = document.querySelectorAll('button') 
expect(allButtons).toBeDisabled(); // Should this pass?
expect(allButtons).not.toBeDisabled(); // And this?
```

However, I'm not sure that's what should happen. Perhaps these should all throw an exception when they get an empty collection of elements?

kentcdodds:
Thinking about what people are probably _trying_ to do, I think that an empty list should throw an error in either case. If they want to verify the list is empty then they can add an assertion for that specifically. I think doing anything else could lead to passing tests but a broken app.

gnapse:
Exactly what I was thinking.

smacpherson64:
Sounds good! I will aim to take a stab at implementing the idea this week!

smacpherson64:
@gnapse and @kentcdodds, I have an HOF concept but want to make sure it is on the right track.
My thought is the HOF wrapping each matcher in index.js would allow us to not have to adjust any of the matchers directly. If the HOF determines that it is not a nodelist it just runs the normal matcher.

https://github.com/smacpherson64/jest-dom/commit/17770e25c5c0a4fece84bc4b7bcfbcf4c9d0fe4d

There is a lot that needs to be cleaned up. Feedback, negative or positive, is appreciated! Thanks for letting me be part!

gnapse:
On a second thought, I'm a bit worried about the complexity this feature introduces in the code, compared to the return. In part I'm thinking of this now, after seeing the code, and realizing that this will make error messages more complex and difficult to be clear enough.

[...]

PR #54, opened 2018-09-03
feat: extend matchers for node lists

smacpherson64:
**What**:
Extending matchers to allow the passing of NodeLists. (#52)

**Why**:
This allows users  to not have to use loops over node lists when testing multiple elements.

**How**:
Created an HOF that wraps each matcher, if the first argument is a NodeList then it runs the matcher on all 

**Checklist**:
* [ ] Documentation
* [ ] Tests
* [ ] Ready to be merged

**NOTE: This is the initial concept and may require tweaking, likely not ready to be merged.**


gnapse:
I had forgotten about this. And to be honest, I'm very sorry to say that I'm not as convinced we need it, as I was before.

However, I'll take a look at how it behaves. I'm mostly worried about how will failing matchers convey the information of why it failed, given the potential of failing due to a lot of elements. That coped with the fact that this is merely a convenience for something users can do themselves relatively easily, makes me wonder.

I know we discussed this issue of conveying failed matches, and I'm sure you devoted significant effort in trying to make it usable. Just let me take a look at that, and at how this adds complexity to the source code. Hopefully, from your description above, this should not be something affecting the code of all matchers. This is something that would wrap any matcher meant for a single node, and inject the magic.

In the mean time, if you can help clarify some of these concerns, or argue why we need it, please do so.

smacpherson64:
Hey @gnapse, 

> I had forgotten about this. And to be honest, I'm very sorry to say that I'm not as convinced we need it, as I was before. 
>
> However, I'll take a look at how it behaves. I'm mostly worried about how will failing matchers convey the information of why it failed, given the potential of failing due to a lot of elements. That coped with the fact that this is merely a convenience for something users can do themselves relatively easily, makes me wonder.

Sounds good, even if we don't end up merging it in, it was fun to build! I think the display is pretty helpful and gives a little more context when an error occurs for a NodeList.

> I know we discussed this issue of conveying failed matches, and I'm sure you devoted significant effort in trying to make it usable. Just let me take a look at that, and at how this adds complexity to the source code. Hopefully, from your description above, this should not be something affecting the code of all matchers. This is something that would wrap any matcher meant for a single node, and inject the magic.

Awesome, Thank you! please take a look! It does wrap the elements so the matchers themselves should not be affected. 

> In the mean time, if you can help clarify some of these concerns, or argue why we need it, please do so.

Makes sense, I will do my best to display a contrast:

The code itself for a simple example does not change significantly but is reduced for the end user. A more complex example, wanting to know how many elements failed or what the path is, would require a bit more on the end users part, but I assume the typical example would be like the following: 

**Before:**
```javascript
const elements = document.querySelectorAll('div');

for (element of elements.values()) {
    expect(element).someMatcher(somevalues);
}
```

**After**
```javascript
const elements = document.querySelectorAll('div');
expect(elements).someMatcher(somevalues);
```

The message that a user would receive does change a bit, the before version stops at the first error. The after version checks all elements and displays more context: 
- How many elements failed / passed
- Displays a sample of the elements and where they are located.

If a user is looping through elements in a NodeList and uses a selector that is too generic: 'div' currently they would need to dig or rely on the matcher to figure out what element the error occurred on. With the NodeList version it has the potential to improve that experience, because a path is displayed. 

**Before:**
```
{matcher error}
```

**After:**
```
# of # elements failed

[index]: {shallow clone}
> path to element 

[index]: {shallow clone}
> path to element 

[index]: {shallow clone}
> path to element 

... (stops the sample at 3)

Error message for element [0]: 
============================
{matcher error}
============================
```

PR #63, opened 2018-10-06
fix: upgrade jest-diff and jest-matcher-utils

SimenB:
**What**:

Upgrade dated dependencies.

**Why**:

Mostly to dedupe better on new projects, but it also brings in a bugfix (commented inline)

**How**:

`npx npm-check-updates -u`.

**Checklist**:

* [ ] Documentation N/A
* [ ] Tests N/A
* [x] Ready to be merged 

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,13 +34,13 @@
   "dependencies": {
     "chalk": "^2.4.1",
     "css": "^2.2.3",
-    "jest-diff": "^22.4.3",
-    "jest-matcher-utils": "^22.4.3",
-    "pretty-format": "^23.0.1",
+    "jest-diff": "^23.6.0",
+    "jest-matcher-utils": "^23.6.0",
+    "pretty-format": "^23.6.0",
     "redent": "^2.0.0"
   },
   "devDependencies": {
-    "kcd-scripts": "^0.37.0"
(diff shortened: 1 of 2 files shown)


src/to-have-style.js
```diff
@@ -25,7 +25,6 @@ function printoutStyles(styles) {
     .sort()
     .map(prop => `${prop}: ${styles[prop]};`)
     .join('\n')
-    .concat('\n') // ensure multiline for diff
```
SimenB: Fixed in 23.0.0: https://github.com/facebook/jest/pull/6221

package.json
```diff
@@ -34,13 +34,13 @@
   "dependencies": {
     "chalk": "^2.4.1",
     "css": "^2.2.3",
-    "jest-diff": "^22.4.3",
-    "jest-matcher-utils": "^22.4.3",
-    "pretty-format": "^23.0.1",
+    "jest-diff": "^23.6.0",
+    "jest-matcher-utils": "^23.6.0",
+    "pretty-format": "^23.6.0",
```
SimenB: Not needed as it's within semver range (and `jest-diff` depends on it)m but doesn't hurt to keep them in sync

smacpherson64:
Awesome, Thank you!

PR #64, opened 2018-10-08
Allow running in node environment

### LINKED ISSUE #58: HtmlElementTypeError for custom jsdom instance
* `jest-dom` version: 1.11.0
* `node` version: 10.4.1
* `npm` (or `yarn`) version: yarn 1.7.0

### Relevant code or config:

```js
describe('Index', () => {
  it('Suggests cities to search by', async () => {
    const { window } = await next.render('/')
    expect(window.document).toHaveTextContent('Warszawa')
  })
})
```

next.render is a helper I've created that returns custom jsdom instance

### What you did:

Tried to use toHaveTextContent matcher with custom jsdom instance

### What happened:

Error is thrown:

```
    Received:
      object:
```

### Reproduction:

As above

### Problem description:

It seems you throw HtmlElementTypeError with `htmlElement instanceof HTMLElement` but the element that I pass has separate of HTMLElement created along separate window object.

See:

```
const { window } = new JSDOM(...)
console.log(window.document instanceof Document) # => false
console.log(window.document instanceof window.Document) # => true
```

### Suggested solution:

- Search for Document class in scope of passed container?
- Use `constructor.name`, `__proto__.constructor.name`,` __proto__.__proto__.constructor.name` etc. instead of `instanceof`?

sheerun:
Fixes #58

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -34,8 +34,9 @@ class HtmlElementTypeError extends Error {
 
 function checkHtmlElement(htmlElement, ...args) {
   if (
-    !(htmlElement instanceof HTMLElement) &&
-    !(htmlElement instanceof SVGElement)
+    !htmlElement.ownerDocument &&
+    !(htmlElement instanceof htmlElement.ownerDocument.HTMLElement) &&
+    !(htmlElement instanceof htmlElement.ownerDocument.SVGElement)
   ) {
     throw new HtmlElementTypeError(htmlElement, ...args)
   }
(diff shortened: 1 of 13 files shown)


gnapse:
This is awesome. Thanks! 👏👏👏

src/utils.js
```diff
@@ -34,8 +34,9 @@ class HtmlElementTypeError extends Error {
 
 function checkHtmlElement(htmlElement, ...args) {
   if (
-    !(htmlElement instanceof HTMLElement) &&
-    !(htmlElement instanceof SVGElement)
+    !htmlElement.ownerDocument &&
+    !(htmlElement instanceof htmlElement.ownerDocument.HTMLElement) &&
+    !(htmlElement instanceof htmlElement.ownerDocument.SVGElement)
```
gnapse: Should this condition be this instead?

```js
  if (
    !htmlElement.ownerDocument ||
    (
      !(htmlElement instanceof htmlElement.ownerDocument.HTMLElement) &&
      !(htmlElement instanceof htmlElement.ownerDocument.SVGElement)
    )
  ) {
    throw new HtmlElementTypeError(htmlElement, matcherFn, context, validTypes)
  }
```

If there's no `htmlElement.ownerDocument`, it makes no sense to go and evaluate the rest of the condition, because the rest of the conditional depends entirely on `htmlElement.ownerDocument` not being undefined.

Please, let me know if I'm correct or not so far. If I am, I have more things to say, but I wanted to confirm up to this point at least

cc @sheerun

gnapse: Ok, here it is the additional part of what I wanted to say.

I think the above is never surfaced because `checkHtmlElement` is coincidentally always called with a valid html element that has an `ownerDocument`.

So we never really get to the point where the `instanceof` parts of the conditions are evaluated even. The thing is, when I fixed the condition locally, I get that `htmlElement.ownerDocument.HTMLElement` is undefined. The same with `SVGElement`.

This is the point where I'm no longer sure how to continue this fix. Care to chime in @sheerun?

gnapse:
@sheerun any comments on my concerns above?

PR #65, opened 2018-10-09
Initial toHaveFormValues implementation (closes #60)

### LINKED ISSUE #60: Descriptively assert that a form input/select/textarea has a given value
### The problem

`<select>` elements, unlike `input` elements, do not have a `value` attribute. So there's easy no natural way to assert that a select element has a certain option selected. One would have to traverse through the options to see which one is `selected`.

To make it even worse, React makes it look like so, to the point that I got confused myself recently, when I spent more minutes that I would've liked trying to figure out why the following test (using react-testing-library) did not pass:

```jsx
const { getByTitle } = render(
  <select title="Client status" value="archived">
    <option value="active">Active</option>
    <option value="archived">Archived</option>
    <option value="pending">Pending</option>
  </select>
);

expect(getByTitle('Client status')).toHaveAttribute('value', 'archived')
```

And to continue to make it even worse, the overall task may become more complex if using options grouped in `<optgroup>`s, or with multi-select elements.

### Describe the feature you'd like

I started this issue thinking of suggesting a specific matcher to check a selected option in select elements. But while describing all of the above, I realize that it may be better to provide some generic way to check the value of form elements, regardless of them being `input`s, `textarea`s or `select`s (are there any other ones?).

### Suggested implementation:

It could be providing assertions to work at the actual form input element level:

```js
expect(userInputElement).toHaveValue('janedoe');
expect(passwordInputElement).toHaveValue('12345678');
```

or even at the form element level (identifying an input/select/textarea element by its `name` attribute):

```js
expect(formElement).toHaveValue('username', 'janedoe');
expect(formElement).toHaveValue('password', '12345678');
```

It could be both use cases with the same matcher! And it could be created so that it would understand inputs, selects and textures (which are also different, their value is their child text content).

It could even be so that `<select multiple>` elements' values are arrays, and users could check at once the entire multiple selection (and maybe even part of it too??)

`select` elements probably have some property at the dom level to check their value, instead of it being an attribute. I did not google for it, and do not have it at the top of my head, but if so, then `expect(selectElement.value).toEqual(...)` could help, but still it's not a generic solution for all kinds of form elements (or is it?)

`dom-testing-library` recently added the ability to `getBySelectText`, that allows users to select a `select` (pun not intended) by the text content of the option element within that select that is currently selected. This certainly helps, but it still leaves a weird not-so-expressive way to assert that a given select has a certain value selected:

```js
expect(getByLabel('Client status')).toBe(getBySelectText('Archived'))
```

And even then, what happens if there are two selects that happen to have an option selected, both with the same text `"Archived"`?

And still this only addresses the problem with select elements.

Thoughts?

gnapse:
**What**:

Allow tests to assert that a form (or part of it) contains certain elements with certain corresponding values.

**Why**:

See #60 

**How**:

By gathering all form values under the given container, and matching it against the expected values.

Additional efforts were made to ensure a consistent gathering of form values to match more what a user interprets out of them.

- Handles single vs multiple selection `<select>`
- A set of radio buttons with the same name behave as a single select
- A set of checkboxes with the same name behave as a multi select
- `<input type="number">` is matched as a numeric value, even though internally is a string
- Several elements with the same name are grouped as an array value

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged

### DIFF
--- a/extend-expect.d.ts
+++ b/extend-expect.d.ts
@@ -13,7 +13,7 @@ declare namespace jest {
     toHaveAttribute(attr: string, value?: string): R
     toHaveClass(...classNames: string[]): R
     toHaveFocus(): R
-    toHaveFormValues(object)
+    toHaveFormValues(expectedValues: {[name: string]: any}): R
     toHaveStyle(css: string): R
     toHaveTextContent(
       text: string | RegExp,
--- a/extend-expect.d.ts
+++ b/extend-expect.d.ts
@@ -12,11 +12,12 @@ declare namespace jest {
     toContainHTML(htmlText: string): R
     toHaveAttribute(attr: string, value?: string): R
     toHaveClass(...classNames: string[]): R
+    toHaveFocus(): R
+    toHaveFormValues(object)
     toHaveStyle(css: string): R
     toHaveTextContent(
       text: string | RegExp,
       options?: {normalizeWhitespace: boolean},
     ): R
-    toHaveFocus(): R
   }
 }
--- a/src/to-have-form-values.js
+++ b/src/to-have-form-values.js
@@ -93,9 +93,7 @@ function getPureName(name) {
 }
 
 function getAllFormValues(container) {
-  const names = uniq(
-    [...container.querySelectorAll('[name]')].map(element => element.name),
-  )
+  const names = Array.from(container.elements).map(element => element.name)
   return names.reduce(
     (obj, name) => ({
       ...obj,
--- a/src/to-have-form-values.js
+++ b/src/to-have-form-values.js
@@ -1,17 +1,15 @@
 import {matcherHint} from 'jest-matcher-utils'
 import jestDiff from 'jest-diff'
+import isEqual from 'lodash/isEqual'
+import isEqualWith from 'lodash/isEqualWith'
+import uniq from 'lodash/uniq'
 import {checkHtmlElement} from './utils'
 
-function uniq(value, index, self) {
-  return self.indexOf(value) === index
-}
-
-function isEqual(a, b) {
+function compareArraysAsSet(a, b) {
   if (Array.isArray(a) && Array.isArray(b)) {
-    a = [...a].sort()
-    b = [...b].sort()
+    return isEqual(new Set(a), new Set(b))
   }
-  return JSON.stringify(a) === JSON.stringify(b)
+  return undefined
 }
 
 function getSelectValue({multiple, selectedOptions}) {
@@ -54,7 +52,7 @@ function getSingleElementValue(element) {
 // Returns the combined value of several elements that have the same name
 // e.g. radio buttons or groups of checkboxes
 function getMultiElementValue(elements) {
-  const types = elements.map(element => element.type).filter(uniq)
+  const types = uniq(elements.map(element => element.type))
   if (types.length !== 1) {
     throw new Error(
       'Multiple form elements with the same name must be of the same type',
@@ -95,9 +93,9 @@ function getPureName(name) {
 }
 
 function getAllFormValues(container) {
-  const names = [...container.querySelectorAll('[name]')]
-    .map(element => element.name)
-    .filter(uniq)
+  const names = uniq(
+    [...container.querySelectorAll('[name]')].map(element => element.name),
+  )
   return names.reduce(
     (obj, name) => ({
       ...obj,
@@ -111,8 +109,8 @@ export function toHaveFormValues(formElement, expectedValues) {
   checkHtmlElement(formElement, toHaveFormValues, this)
   const formValues = getAllFormValues(formElement)
   return {
-    pass: Object.entries(expectedValues).every(([name, value]) =>
-      isEqual(formValues[name], value),
+    pass: Object.entries(expectedValues).every(([name, expectedValue]) =>
+      isEqualWith(formValues[name], expectedValue, compareArraysAsSet),
     ),
     message: () => {
       const to = this.isNot ? 'not to' : 'to'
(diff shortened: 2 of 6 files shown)


gnapse:
This is WIP. I still have to make it so it works within a certain form, and not other types of elements. It also lacks documentation and type definitions. But it's ready to be reviewed, so let me know if there's any feedback about the implementation.

src/to-have-form-values.js
```diff
@@ -0,0 +1,130 @@
+import {matcherHint} from 'jest-matcher-utils'
+import jestDiff from 'jest-diff'
+import {checkHtmlElement} from './utils'
+
+function uniq(value, index, self) {
+  return self.indexOf(value) === index
+}
+
+function isEqual(a, b) {
```
kentcdodds: Maybe we should just use `lodash.isequal`

src/to-have-form-values.js
```diff
@@ -0,0 +1,130 @@
...
+  }
+}
+
+// Strips the `[]` suffix off a form value name
+function getPureName(name) {
+  return /\[\]$/.test(name) ? name.slice(0, -2) : name
+}
+
+function getAllFormValues(container) {
+  const names = [...container.querySelectorAll('[name]')]
```
kentcdodds: Rather than this query selector, why don't we rely more on the DOM's built-in mechanism for form elements? This would have the added benefit of not working if the form controls are not wired up to the form properly.

So rather than `container.querySelectorAll('[name]')` where `container` is actually `form`, we could just do: `form.elements`?

src/to-have-form-values.js
gnapse: Yes, I initially resisted to add lodash as a dependency, but as I needed more and more, I considered it.

src/to-have-form-values.js
gnapse: Definitely. As stated in the description, I've yet to check that the container is a form element or a field set.

src/to-have-form-values.js
kentcdodds: I often resist installing lodash, but then I remind myself that the chances someone doesn't already have lodash installed in their `node_modules` are very very slim 😅

gnapse: However, there's still something else being done here besides a mere `isEqual`. Because we need to compare arrays as sets, not as arrays. Order is not important. But anyway, lodash will surely help with that anyway. This was indeed a sort of hack to get this going.

gnapse:
Fixed the two pending observations (use of lodash and form.elements instead of rolling our own solution for those needs).

Still pending to add documentation and type definitions.

Also, there's an issue in #64 that is preventing me from actually checking if the element is an `instance HTMLFormElement` or of `HTMLFieldSetElement`. I added a less reliable test for this for the time being, until that issue is resolved.

gnapse:
Added documentation and type definition.

Not sure if the documentation is too large, but I felt I had to document all cases that this matcher covers. Regardless, feedback on the documentation would be great.

I also could not give a more accurate type definition. I would've liked to be more precise with something like this:

```typescript
toHaveFormValues({
  [name: string]: any
})
```

But for some reason I keep getting a TypeScript syntax error, when in other contexts I've been able to use this just fine. But I figure `toHaveFormValues(object)` is fine too.

Finally, I wonder if it makes sense to validate and error out if the given object is empty. It makes little sense to call `toHaveFormValues({})` but right now it will always pass. Guess that's no big deal either.

extend-expect.d.ts
```diff
@@ -12,11 +12,12 @@ declare namespace jest {
     toContainHTML(htmlText: string): R
     toHaveAttribute(attr: string, value?: string): R
     toHaveClass(...classNames: string[]): R
+    toHaveFocus(): R
+    toHaveFormValues(object)
```
jgoz: @gnapse I think the reason you were seeing syntax errors is because the parameter needs to be named. As it stands, this is also incorrect syntax.

```ts
toHaveFormValues(values: { [name: string]: any })
```

gnapse: Thanks, I knew I must have been doing something wrong. I guess I need to take a break ;)

jgoz: Also, it should specify a return value of `R`

gnapse:
A use case I wanted to bring to others' attention: what happens if there are multiple elements with the same name (other than radio buttons or checkboxes)? How should we treat those?

Right now I considered grouping them all as a single form control that returns a value that is an array of all the values of all the elements with that name. I only y do this if they're also all of the same `type`.

But I wonder if even if they're of the same type, I should do something different, such as failing with an error, or just defaulting to returning the value of only one of them, say the first or the last one. I think this is what submitting such a form would regularly do. But I also feel that we shouldn't be silent about it.

Thoughts?

kentcdodds:
What would happen in an old-school application that uses a form POST of the data? My guess is the later form control wins and it doesn't convert to an array. Could you validate that and let's just do what the browser would do.

gnapse:
In any case I think this matcher is a functionality that will probably get a lot of feedback in form of issues, bug reports, feature requests, after we release it, especially around how it treats edge or unconventional cases.

gnapse:
> What would happen in an old-school application that uses a form POST of the data? My guess is the later form control wins and it doesn't convert to an array. Could you validate that and let's just do what the browser would do.

I will do that. That's what I was inclined to do after thinking about this after I implemented the current approach.

I think I'll also print out a warning of some sort. Not making the matcher fail if there really isn't a mismatch. But also not let this unconventional condition pass through unnoticed. This is not a valid use case after all, and users may get failing tests inadvertently because of this.

kentcdodds:
If there's a log, there should be a way to disable the log 👍

gnapse:
Update: it seems to be valid to have multiple elements with the same name, not just radio buttons and checkboxes. Moreover, they seem to interpreted as an array, so I guess the current implementation is ok. Source: https://stackoverflow.com/questions/2906793/is-it-valid-to-have-two-input-elements-with-the-same-name

If agreed, I think this is ready to be reviewed and released if ok.

PR #69, opened 2018-10-31
fix: minor typos in README

bewildergeist:
**What**:
Minor adjustments to the README:
- A typo in a variable name
- A few copy-paste errors in variable names
- Changing the order of a couple of CSS properties to make it clear that the assertion fails because of a different value, not a different order of properties

**Why**:
Just makes the docs easier to read.

**How**:
N/A

**Checklist**:
* [x] Documentation
* [ ] Tests N/A
* [x] Ready to be merged
* [ ] Added myself to contributors table N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,8 +199,8 @@ const detachedElement = document.createElement('div')
 
 expect(htmlElement).toBeInTheDocument()
 expect(svgElement).toBeInTheDocument()
-expect(detachedElement).not.toBeInTheDocument()
 expect(nonExistantElement).not.toBeInTheDocument()
+expect(detachedElement).not.toBeInTheDocument()
 ```
 
 ##### Using dom-testing-library
--- a/README.md
+++ b/README.md
@@ -199,7 +199,7 @@ const detachedElement = document.createElement('div')
 
 expect(htmlElement).toBeInTheDocument()
 expect(svgElement).toBeInTheDocument()
-expect(detacthedElement).not.toBeInTheDocument()
+expect(detachedElement).not.toBeInTheDocument()
 expect(nonExistantElement).not.toBeInTheDocument()
 ```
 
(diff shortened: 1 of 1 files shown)


gnapse:
Hi, thanks!

One thing though. I do not get why your changes regarding css properties order are consistent with the reason you state here?

> Changing the order of a couple of CSS properties to make it clear that the assertion fails because of a different value, not a different order of properties

The css properties were already in the same order in the sample element as in the sample test code. This made it clear that the order is not important. Because even with the same order, the `.not.toHaveStyle` assertions still pass.

README.md
```diff
@@ -631,7 +631,7 @@ This API allows you to check whether the given element has a text content or not
 ##### Using document.querySelector
 
 ```javascript
-const button = document.querySelector('[data-testid="count-value"]')
+const content = document.querySelector('[data-testid="count-value"]')
```
gnapse: Thanks for spotting this inconsistency. But now that we're at it, the name `content` is not good either. Can you change it to `element` and also use that in the `expect` assertions below?

bewildergeist: Good call — updated in 0b887ac.

bewildergeist:
@gnapse Yeah, the CSS order thing is a very minor niggle, I'm happy to remove that again. My thinking was:

1. The sample element specifies the order `display: none; color: red;`
2. The second assertion specifies that `color: red; display: none;` will be `true`, thus already proving that property order doesn't matter (as anyone who knows CSS will know, of course).
3. Then the third assertion should specify that `color: blue; display: none;` will be `false` and show the properties in the same order as the previous assertion, to make it immediately apparent to the reader that `red`/`blue` is the only difference between those two assertions.

The fact that the order of properties is inconsequential should basically be a given, so I just thought we should lower the "cognitive overhead" of telling those two assertions apart. But as I said, definitely a very minor point, happy to revert ☺️

PR #72, opened 2018-12-03
Fix toHaveTextContent to be case-sensitive

### LINKED ISSUE #71: Should .toHaveTextContent be case-sensitive?
I don't know if this is a bug or it is a feature, but I expected that `.toHaveTextContent` was case-sensitive. I could address that if it is a bug.

* `jest-dom` version: 2.1.1

### What you did:

```html
<div className="App">
  <h1>Hello CodeSandbox</h1>
  <h2>Start editing to see some magic happen!</h2>
</div>
```
I expect the expect below would fail, but it passes.
```js
expect(subject.container.querySelector("h1")).toHaveTextContent("hello codesandbox");
```

lourenci:
**What**:
Change `.toHaveTextContent` to be case-sensitive. It closes #71.

**Why**:
This is the expected usage. If the user wants insensitivity, he should do through `Regexp`.

**How**:
Is was removed the `.toLowerCase` from the `match` utils function.

**Checklist**:

* [ ] Documentation
* [x] Tests
* [x] Ready to be merged 

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -116,7 +116,7 @@ function matches(textToMatch, matcher) {
   if (matcher instanceof RegExp) {
     return matcher.test(textToMatch)
   } else {
-    return textToMatch.includes(matcher)
+    return textToMatch.includes(String(matcher))
   }
 }
 
--- a/src/utils.js
+++ b/src/utils.js
@@ -116,7 +116,7 @@ function matches(textToMatch, matcher) {
   if (matcher instanceof RegExp) {
     return matcher.test(textToMatch)
   } else {
-    return textToMatch.toLowerCase().includes(String(matcher).toLowerCase())
+    return textToMatch.includes(matcher)
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/utils.js
```diff
@@ -116,7 +116,7 @@ function matches(textToMatch, matcher) {
   if (matcher instanceof RegExp) {
     return matcher.test(textToMatch)
   } else {
-    return textToMatch.toLowerCase().includes(String(matcher).toLowerCase())
+    return textToMatch.includes(matcher)
```
gnapse: I would keep `matcher` wrapped in `String()`. This allows to pass non string stuff (such as numbers) to `.toHaveTextContent`, and removing it as part of this change adds more surface for tests to break for issues not related with what you're trying to fix (case sensitiveness).

```suggestion
    return textToMatch.includes(String(matcher))
```

lourenci: It makes sense. I'll do it.

lourenci:
Yes, it's a breaking change. Some tests written believing in insensitivity will break.

PR #77, opened 2019-02-06
Pr/escape css strings

mufasa71:
**What**:

Special characters should be escaped in `querySelectorAll`

**Why**:

Tests are failing due to unescaped sequence in name attribute

**How**:

css.escape library was added

**Checklist**:

* [ ] Documentation
* [x] Tests
* [x] Ready to be merged 


gnapse:
Thanks!

PR #79, opened 2019-02-10
fix: toBeInTheDocument fails on null

dreyks:
**What**:
`expect(null).toBeInTheDocument()` fails with a non-descriptive error `TypeError: Cannot read property 'ownerDocument' of null`

**Why**:
to produce a more friendly error message

**How**:
- when doing a positive expectation do call checkHtmlElement to allow it to fail on `null`
- also made negative cases to be more specific since `toThrowError()` with no arguments passes on any error including logic errors

**Checklist**:

* [ ] Documentation N/A
* [x] Tests
* [x] Ready to be merged 

Alternatively instead of calling checkHtmlElement we can change https://github.com/gnapse/jest-dom/blob/02c340d03f5b9bb5b02c51286f306ddd1759ae42/src/to-be-in-the-document.js#L26 to something like `${stringify(element ? element.ownerDocument.cloneNode(false) : 'document')}`. Tell me if that's preferable (it also makes the error message even more obvious)


gnapse:
Thanks!

PR #85, opened 2019-03-12
Fix toBeVisible ignoring hidden attribute

### LINKED ISSUE #84: .toBeVisible should consider "hidden" attribute

`expect(element).toBeVisible()` should return `false` if `element.hasAttribute('hidden')` unless it's been explicitly shown with styles (`display && display !== 'none'` for example).

### Suggested implementation:

Add `&& !element.hasAttribute('hidden')` here, plus a check on `display`:
https://github.com/gnapse/jest-dom/blob/238cc91bfaffa9dd5baed1b5a02d060909f7c30c/src/to-be-visible.js#L17-L22

I'm using `.toHaveAttribute("hidden")`, but I feel like I'm testing an implementation detail.

diegohaz:
Closes #84

It checks the presence of the `hidden` attribute. Unfortunately, it seems that the case when it's combined with `display !== "none"` can't be handled within `jsdom`. In browsers, we could check `element.offsetParent` and other layout properties, but `jsdom` doesn't support it.

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged 


diegohaz:
Added docs.

Any `display` value other than `none` overrides the behavior of the `hidden` attribute in the browser if explicitly defined. This means that `<div style="display: block" hidden>` is actually visible in the browser.

We can check inline style, but as far as I know, we can't do that with external css. `getComputedStyle().display` is always filled.

Issue #82, opened 2019-03-04
add .toHaveValue matcher


lukaszfiszer:
### The problem

#65 implemented a high-level API for asserting values of an entire form, which is great. However, this API it does not allow to assert the value of a single form element, which is more desired is some scenarios

### Describe the feature you'd like:

Add `toHaveValue` matcher (as described in the initial comment in #60) which can be used to assert value of a single form element - an `input`, `select` or `textarea`

```
expect(getByLabelText('Username')).toHaveValue('janedoe');
```

### Suggested implementation:

Create `toHaveValue(formElement, expectedValues)` that calls existing `getSingleElementValue` (https://github.com/gnapse/jest-dom/blob/master/src/to-have-form-values.js#L38)

I will be glad to help with implementing this feature

gnapse:
Hmmmm, and what about this?

```javascript
toHaveFormValues({ username: 'janedoe' })`
```

Though I would not be opposed to add this extra matcher too. That was actually my first idea for this functionality, but then we realized that by providing the more generic one, users could make the simpler case work too, as depicted above.

lukaszfiszer:
Yeah, this works of course. However, there are scenarios when `expect(element).toHaveValue(value)` may be preferred:

1. When I already have a reference to my field element:

```
const {getByLabel} = render(<Form/>);
const usernameInput = getByLabel('username');
expect(usernameInput).toBeEnabled();
expect(usernameInput).toHaveFocus();
expect(usernameInput).toHaveValue('some_value');
```

2. When I don't care (or don't know) what `name` attribute my field has. I can treat it as an implementation details and reference the field by label or placeholder, just as users do. This fits well into philosophy of react-testing-library, and https://testing-library.com/docs/guiding-principles especially

kentcdodds:
I agree. This would be a great addition 👍

gnapse:
I definitely agree. Go for it. and thanks for your ideas and contributions!

lukaszfiszer:
Great! I will submit a PR soon

lukaszfiszer:
Closed in #90

PR #90, opened 2019-04-01
feat: Add toHaveValue matcher #82

lukaszfiszer:
**What**:
Add `toHaveValue` matcher

**Why**:
As discussed in #82 

**How**:
Reused a lot of existing logic from `toHaveFormValue` and moved some internal functions  to `utils.js` so that they can be shared between both modules.

I had a lot of thoughts on how to implement this matcher for `<input type="checkbox" />` and `<input type="radio" />`, because those elements value can be different depending on other elements from the same group (sharing the same `name`). Finally I've decided to throw an error when a checkbox or radio input is passed, suggesting users to use `toHaveFormValue` instead. In the future, I think I would most make sense to implement `toBeChecked` matcher - a perfect pair with to `toHaveValue` for single form elements. 

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged 

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@
 [![downloads][downloads-badge]][npmtrends]
 [![MIT License][license-badge]][license]
 
-[![All Contributors](https://img.shields.io/badge/all_contributors-25-orange.svg?style=flat-square)](#contributors)
+[![All Contributors](https://img.shields.io/badge/all_contributors-26-orange.svg?style=flat-square)](#contributors)
 [![PRs Welcome][prs-badge]][prs]
 [![Code of Conduct][coc-badge]][coc]
 
@@ -61,6 +61,7 @@ to maintain.
   - [`toHaveFormValues`](#tohaveformvalues)
   - [`toHaveStyle`](#tohavestyle)
   - [`toHaveTextContent`](#tohavetextcontent)
+  - [`toHaveValue`](#tohavevalue)
 - [Deprecated matchers](#deprecated-matchers)
   - [`toBeInTheDOM`](#tobeinthedom)
 - [Inspiration](#inspiration)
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -687,6 +688,63 @@ expect(element).not.toHaveTextContent('content')
...
+### `toHaveValue`
+
+```typescript
+toHaveValue(value: string | string[])
+```
+
+This allows you to check whether the given form element has the specified value.
+It accepts `<input>`, `<select>` and `<textarea>` elements with the exception of
+of `<input type="checkbox">` and `<input type="radio">`, which can be meaningfully
+matched only using [`toHaveFromValue`](#tohaveformvalues).
```
gnapse: Typo `toHaveFromValue` should be `toHaveFormValue`

README.md
```diff
@@ -687,6 +688,63 @@ expect(element).not.toHaveTextContent('content')
...
+
+For all other form elements, the value is matched using the same algorithm
+as in [`toHaveFromValue`](#tohaveformvalues) does.
```
gnapse: Same typo.

lukaszfiszer:
@gnapse I have fix the typos.

Another idea that I had for checkbox and radio inputs was to allow the user to pass a parent element of those elements:

```html
<div data-testid="contact">
   <label><input type="radio" name="contact" value="email">Email</label>
   <label><input type="radio" name="contact" value="phone">Phone</label>
   <label><input type="radio" name="contact" value="post">Post</label>
</div>
```

```js
expect(getByTestId('contact')).toHaveValue('phone')
```

But then probably you could  just use `toHaveFormValues` for that.

Let me know what you think

kentcdodds:
What do we need to do to help get this across the finish line?

gnapse:
I probably forgot to keep track of this and approve or request changes at the time. But now it seems there are conflicts, and also the build is failing for some reason (could it have something to do with the change of url of the github repo when moved to an org @kentcdodds?  When I visit the travis details link above I get "We couldn't find the repository gnapse/jest-dom").

@lukaszfiszer sorry about not being on top of this. Can you please fix the conflicts? We'll figure out the travis issue in the mean time.

lukaszfiszer:
@gnapse No worries. I will fix the conflicts over the weekend.

lukaszfiszer:
@gnapse rebased against master and fixed the broken test. ready to merge

extend-expect.d.ts
```diff
@@ -20,5 +20,6 @@ declare namespace jest {
       text: string | RegExp,
       options?: {normalizeWhitespace: boolean},
     ): R
+    toHaveValue(value?: string | string[]): R
```
jgoz: Should this be `value?: string | string[] | number`?

README.md
```diff
@@ -690,6 +691,63 @@ expect(element).not.toHaveTextContent('content')
 
 <hr />
 
+### `toHaveValue`
+
+```typescript
+toHaveValue(value: string | string[])
```
jgoz: And should this one be `value?: string | string[] | number`?

src/__tests__/to-have-value.js
```diff
@@ -0,0 +1,88 @@
...
+describe('.toHaveValue', () => {
+  test('handles value of text input', () => {
+    const {queryByTestId} = render(`
+        <input type="text" value="foo" data-testid="with" />
+        <input type="text" value=""  data-testid="empty" />
+        <input type="text" data-testid="without" />
+    `)
+
+    expect(queryByTestId('with')).toHaveValue('foo')
+    expect(queryByTestId('with')).not.toHaveValue('bar')
```
jgoz: Would it make sense to add a test case like this?

```js
expect(queryByTestId('with')).toHaveValue();
```

The corresponding `.not.toHaveValue()` case is already there for `empty`.

lukaszfiszer:
@jgoz Thanks for review. I've fixed the typings, added additional test cases and fixed the behaviour of `.toHaveValue()` (without argument) - it asserts if input/select has *any* value.

src/index.js
```diff
@@ -29,7 +30,9 @@ export {
   toBeVisible,
   toBeDisabled,
   toBeEnabled,
+  toHaveValue,
   toBeRequired,
   toBeInvalid,
   toBeValid,
+  toHaveValue
```
jgoz: This is being exported twice

src/to-have-value.js
```diff
@@ -0,0 +1,44 @@
...
+
+export function toHaveValue(htmlElement, expectedValue) {
+  checkHtmlElement(htmlElement, toHaveValue, this)
+
+  if (
+    htmlElement.tagName.toLowerCase() === 'input' &&
+    ['checkbox', 'radio'].includes(htmlElement.type)
+  ) {
+    throw new Error(
+      'input with type=checkbox or type=radio cannot be used with .toHaveValue(). Use .toHaveFormValue() instead',
```
jgoz: Minor: `.toHaveFormValue()` should be `.toHaveFormValues()`

jgoz:
@lukaszfiszer sorry for the delayed response, I missed the notification on this. Just two very minor issues that I can see, one of which is causing the CI failure.

lukaszfiszer:
@jgoz fixed issues and squash-rebased against master. Hope to get it merged!

README.md
```diff
@@ -839,6 +840,63 @@ expect(element).not.toHaveTextContent('content')
...
+expect(emptyInput).not.toHaveValue()
+expect(selectInput).not.toHaveValue(['second', 'third'])
+```
+
+##### Using dom-testing-library
+
+```javascript
+const {getByTestId} = render(/* Rendered HTML */)
+
+const textInput = getByTestId('[data-testid="input-text"]')
```
kentcdodds: I think these should be: `getByTestId('input-text')` and so-forth.

lukaszfiszer: Good catch! Fixed.

Issue #98, opened 2019-04-25
toContainHTML failure output is not useful


ratbeard:
* `jest-dom` version: 3.1.3
* `node` version: v10.11.0
* `npm` (or `yarn`) version: 6.9.0

### What you did:

```js
const container = render(<SubmitButton>)
expect(container).toContainHtml('color: blue')
```

### What happened:

<img width="420" alt=" 2019-04-25 at 3 27 53 PM" src="https://user-images.githubusercontent.com/28356/56766644-83b1a900-676f-11e9-8797-4f396af6e37e.png">

### Problem description:

The output failure message does not contain any useful information for tracking down the reason why it failed.  

I have to convert it to something like `expect(container.innerHTML).toContain('color: blue')` to be able to debug it, making the custom matcher less efficient than just using the simpler builtin option.

Note, I saw theres a style matcher that could be used in this specific test, however the problem applies to any call to this matcher.

### Suggested solution:

Pretty sure it just needs to change to `cloneNode(true)` at https://github.com/testing-library/jest-dom/blob/master/src/to-contain-html.js#L14 for it to print out all the nodes.  

That fix would make it debuggable.  To make it better than the output of `toContain()` it could be pretty-printed as indented and syntax highlighted html 😃 - any chance that has been considered or used by other matchers?  

Here is the toContain() output for reference.  You can imagine its much harder to debug a larger dom node

<img width="597" alt=" 2019-04-25 at 3 40 59 PM" src="https://user-images.githubusercontent.com/28356/56767258-e0619380-6770-11e9-8601-04883fe82460.png">

gnapse:
Thanks for reporting this.

If you want to assert that an element has some style, you should prefer using [`.toHaveStyle`](https://github.com/testing-library/jest-dom#tohavestyle). The matcher you're trying to use is not meant for that. I know you mentioned it, but I wanted to stress that point, because fixing this should not be something encourage checking styles in that way.

In addition to not being intended to test for inline css styles, `toContainHTML` is not even really recommended in the general case, if you know the structure of the html that you want to test, you better test it in a more semantical way. See this disclaimer also found in the README about `.toContainHTML`:

> It could be useful in situations where the code being tested renders html that was obtained from an external source, and you want to validate that that html code was used as intended.

However, with all that being said, I do agree with you that if switching that line of code would give better errors, then we should do it. Can you contribute it? Otherwise, I may get to it at some point in the next few days, if it's really that simple.

[...]

PR #100, opened 2019-04-29
fix: display html in toContainHTML failure output #98

ratbeard:
**What**:

#98 prints out html when toContainHTML matcher fails

**Why**:

Before:
2019-04-29 at 12 57 51 PM

After:
2019-04-29 at 12 57 36 PM

**How**:

**Checklist**:

* [] Documentation n/a
* [] Tests n/a
* [ ] Updated Type Definitions n/a
* [x] Ready to be merged 


rikkit:
To match the behaviour of other matchers it would be nice if this also printed the value of `expected`. Would a PR like that be accepted?

gnapse:
@rikkit yes, go for it. Thanks!

PR #93, opened 2019-04-15
feat(toHaveAttribute): support asymmetric matchers for value

revathskumar:
Fixes #91

**Sample Error message**
Screenshot from 2019-04-14 19-47-15

**What**:

`toHaveAttribute` value will accept `asymmetric` matchers

**Why**:

`asymmetric` matchers are handy when we want to do partial matching

**How**:

Using the method [equals](https://github.com/facebook/jest/blob/ac8c345f5318016d0789974fbc815d857fd70d99/packages/expect/src/jasmineUtils.ts#L30) in `expect/JasmineUtils`

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Updated Type Definitions
* [ ] Ready to be merged 
* [x] Added myself to contributors table

PR #103, opened 2019-05-08
fix: use `this.equals` instead of import jasmineUtils

revathskumar:
**What**:

As per the [comment](https://github.com/testing-library/jest-dom/pull/93#discussion_r282040908) by @SimenB

This will remove the direct import of jasmineUtils

`import {equals} from 'expect/build/jasmineUtils'`

and instead use `this.equals`.

Also removes the added dependency of `expect`.

**Why**:

**How**:

**Checklist**:

* [ ] Documentation
* [ ] Tests
* [ ] Ready to be merged 

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,6 @@
     "chalk": "^2.4.1",
     "css": "^2.2.3",
     "css.escape": "^1.5.1",
-    "expect": "^24.7.1",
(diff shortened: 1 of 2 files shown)


SimenB:
Perfect! 🙂

Issue #104, opened 2019-05-10
toHaveTextContent("") matches element with non-empty content


alexkrolick:
v3.2.2

```js
const el = document.createElement("div")
el.textContent = "0"
expect(el).toHaveTextContent("") // passes, should be error
```

alexkrolick:
Will add a repro test case, the example I tried passed (unexpectedly)

gnapse:
Is this happening only with the string `"0"`? Because from the code I think any string will match `.toHaveTextContent('')`. This is all ultimately determined [here](https://github.com/testing-library/jest-dom/blob/b2b4bd3ba65c48b350194a7623797e70f8a1390e/src/utils.js#L119), and any string `.includes('')`.

alexkrolick:
Ah, indeed, it fails to throw for other strings.

alexkrolick:
I don't think it should use `includes` here - use a regex for partial matches, or something like `toContainText`

PR #105, opened 2019-05-10
test: add failing test for #104

alexkrolick:
#104 reproduction (WIP)


gnapse:
@alexkrolick can we close this? Seems #104 was closed by the fix introduced in #138, right?

Issue #108, opened 2019-05-21
Add .toBeInvalid and .toBeRequired


hiwelo:
### Describe the feature you'd like:

I was wondering if we could add `toBeRequired` and `toBeInvalid` as tests based on the same logic as `toBeDisabled`.

The **toBeRequired** test could be based on either having:

- a `required` attribute;
- an `aria-required` attribute.

The **toBeInvalid** test could be based on having:

- an `aria-invalid` attribute.

For accessibility reasons, I am writing a lot of tests using these requirements and it would be way cleaner if I could use the same kind of syntax as `toBeDisabled`.
What are your thoughts about that?

### Describe alternatives you've considered:

At the moment, I am using `.toHaveAttribute('aria-invalid')` or similar.

hiwelo:
Just one more information, this would be really interesting for me especially for the `.toBeRequired` because it could avoid having two different lines for the same test. 👍

gnapse:
This sounds interesting.

Although I'm a bit skeptical in general given the simplicity of the alternative, especially in the case of the `.toBeInvalid`, where you can do this with a single line. However, on a deeper check of the documentation of this attribute, it's not as simple as checking if the attribute is present, because it could be present with a value of `false`, which means the element is not invalid.

And as you said, `.toBeRequired` is a bit more justified, given that it needs to check for the presence of two different attributes. Also in this case, `aria-required` can be `false`, which means is not required.

So overall, yes, I think this would be a good addition to the set of matchers in this library. I'll link here to the spec of these attributes so that all the rules are kept in mind when developing this:

- [aria-invalid](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-invalid_attribute)
- [aria-required](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-required_attribute)
- [required](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-required) (this one has quite a few caveats where the attribute does not apply or does not have the intended effect)

Another thing to keep in mind is what's the precedence of `required` and `aria-required`? What happens if both are present and indicating opposite things? (regardless of which comes first, we should issue a warning in those cases).

[...]

PR #109, opened 2019-05-21
Add toBeRequired() method

hiwelo:
**What**:
This pull request is adding `.toBeRequired()` ~and `.toBeOptional()` methods~ allowing folks to test if a form element is ~optional or~ required.

These methods are testing if the form element is having either `aria-required` or `required` attributes.
In the case of a conflict between these two attributes, the value of `required` is having precedence.

**Why**:

This change is following the discussion within the issue #108.

**How**:

I added a new file in `src/` with the different tests inspired from `to-be-disabled.js`.
I added the script to `src/index.js`.
I added a test in `src/__tests__/`.

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged 


src/to-be-required.js
```diff
@@ -0,0 +1,68 @@
...
+  return element.tagName && element.tagName.toLowerCase()
+}
+
+function isElementRequired(element) {
+  return FORM_TAGS.includes(getTag(element)) && element.hasAttribute('required')
+}
+
+function isElementRequiredByARIA(element) {
+  return (
+    FORM_TAGS.includes(getTag(element)) &&
```
gnapse: On the other hand I suspect that `aria-selected` is not as restricted, and may be usable even in `div` elements with certain `role`s ([ref](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-required_attribute#Used_in_ARIA_roles)). Though again, I'm not suggesting we go as far as ensuring that if present the element has one of these roles (unless you think is easily doable and makes sense). But at least we should probably not enforce that the element tag is one of these.

src/to-be-required.js
```diff
@@ -0,0 +1,68 @@
+import {matcherHint, printReceived} from 'jest-matcher-utils'
+import {checkHtmlElement} from './utils'
+
+// form elements that support 'required'
```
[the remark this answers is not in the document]
hiwelo: Yes, thanks for spotting. I will update and I will also have a check to see if the field is read-only. And if it's the case, we are not checking either.

src/to-be-required.js
hiwelo: We could enforce that the attribute is present on either:

* a block having `role="combobox"`, `role="gridcell"`, `role="radiogroup"`, `role="spinbutton"` or `role="tree"`
* a block having or a form element corresponding to the `role="listbox"` (so any `<selectbox>`)
* an block having or a form element corresponding to the `role="textbox"` (so any `<input>` and `<textarea>`)

hiwelo:
Also, I wonder if I will not remove `.toBeOptional`. It is just a reverse check but if with the full ARIA implementation I am not sure it makes still sense. Because the lack of `required` or `aria-required="true"` does not mean it is optional. An form requirement enforcement can still occur server-side. What are your thoughts on this @gnapse?

hiwelo:
@gnapse New commits updating the PR with the requested changes.
I also added the different scenarios in the tests and the documentation.

Also, I checked the code coverage & added the type definition.

hiwelo:
@gnapse I just pushed the requested changes. Basically, I removed `.toBeOptional` from:

* the initial script file
* the README.md file
* the type declaration
* the jest tests

Please let me know if something else is needed :+1:

hiwelo:
I just tried to force-push an amended commit to trigger the Travis checks but looks like it is just not working 🙂

PR #110, opened 2019-05-23
Implement `toBeInvalid` & `toBeValid` matchers

hiwelo:
**What**:

This pull-request is a proposed implementation for the custom matchers `toBeInvalid` and `toBeValid`.

**Why**:

This is following a discussion we quickly had with the issue #108.
These two custom matchers can avoid folks to check with different expect tests:

- is having `aria-invalid` not equal to `false`
- `checkValidity` is returning `false`

**How**:

* `toBeInvalid` checks if one of these are correct:

  * `aria-invalid` exists and if the value is not equal to `false` (as each value different than false are considered true ([ref](https://www.w3.org/TR/wai-aria-1.1/#aria-invalid)))

  * the targetted element is a form element and `checkValidity()` returns `false`

* `toBeValid` checks if all of these are correct:

  * `aria-invalid` is not part of the markup or the value is equal to `false`

  * if the targetted element is a form element, `checkValidity()` returns `true`

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged 

I do not know if it's a good practice, please let me know your thoughts about that: I had to use `jsdom` to be able to test if these two custom matchers are correctly checking `.checkValidity()`.
Without it, `.checkValidity()` was always returning `true` even when the `input` had an invalid value.

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,7 +245,7 @@ toBeInvalid()
 
 This allows you to check if an form element is currently invalid.
 
-An element is invalid if it is having an `aria-invalid` attribute or if the result of `checkValidity()` are false.
+An element is invalid if it is having an [`aria-invalid` attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-invalid_attribute) or if the result of [`checkValidity()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/checkValidity) are false.
 
 #### Examples
 
@@ -352,7 +352,7 @@ toBeValid()
 
 This allows you to check if the value of a form element is currently valid.
 
-An element is valid if it is not having an `aria-invalid` attribute or having `false` as a value and returning `true` when calling `checkValidity()`.
+An element is valid if it is not having an [`aria-invalid` attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-invalid_attribute) or having `false` as a value and returning `true` when calling [`checkValidity()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/checkValidity).
 
 #### Examples
 
--- a/README.md
+++ b/README.md
@@ -49,7 +49,9 @@ to maintain.
   - [`toBeEnabled`](#tobeenabled)
   - [`toBeEmpty`](#tobeempty)
   - [`toBeInTheDocument`](#tobeinthedocument)
+  - [`toBeInvalid`](#tobeinvalid)
   - [`toBeRequired`](#toberequired)
+  - [`toBeValid`](#tobevalid)
   - [`toBeVisible`](#tobevisible)
   - [`toContainElement`](#tocontainelement)
   - [`toContainHTML`](#tocontainhtml)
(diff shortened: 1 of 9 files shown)


src/__tests__/to-be-invalid.js
```diff
@@ -0,0 +1,72 @@
...
+  const {queryByTestId} = render(`
+    <div>
+      <input data-testid="no-aria-invalid">
+      <input data-testid="aria-invalid" aria-invalid>
+      <input data-testid="aria-invalid-value" aria-invalid="true">
+      <input data-testid="aria-invalid-false" aria-invalid="false">
+    </div>
+    `)
+
+  const dom = new JSDOM(`<input pattern="[a-z]{1,3}" value="test+">`)
```
gnapse: I got your comment about this. I do not like it too much, but if putting this above did not work we're have to live with it. Maybe add a one-liner comment above this line with a `TODO:` and a link to the text where you explained this in your PR, or an explanation here, for future context if we can ever fix it.

src/__tests__/to-be-invalid.js
```diff
@@ -0,0 +1,72 @@
...
```
gnapse: Same comment here. Or probably extract this into a helper function in this very file where you can document the issue just once.

One problem I see with this is that right now tests are runnable in a browser environment. I would be open to merge this and wait to see if this breaks something for someone. At most this would break running the tests for `jest-dom`, not the usage of `jest-dom` itself in other people's tests, right?

I also have an idea based on this. Maybe we can change the implementation of the `render` helper to not rely on `document.innerHTML = ...`, which seems to be part of the problem.

README.md
```diff
@@ -235,6 +237,45 @@ expect(
...
 
+### `toBeInvalid`
+
+```typescript
+toBeInvalid()
+```
+
+This allows you to check if an form element is currently invalid.
+
+An element is invalid if it is having an `aria-invalid` attribute or if the result of `checkValidity()` are false.
```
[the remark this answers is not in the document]
hiwelo: I will add this, I was more thinking about this page: https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/checkValidity.

Also, this page is linking yours as a resources :+1:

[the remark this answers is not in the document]
hiwelo: After a bit more reading, I used your link. It is more generic. My link was indeed about `checkValidity` but in the context of the HTML `select` element. Valid, but I would prefer something more generic.

package.json
```diff
@@ -43,7 +43,7 @@
     "redent": "^2.0.0"
   },
   "devDependencies": {
-    "jsdom": "^12.2.0",
```
hiwelo: @gnapse In fact, I am just realizing that the jsdom library was already part of this project.
I am not sure I would need this update. Will rollback to 12.2 and seeing if it's still working.

If it's the case, I should be able to remove the "Warning" comment in the test file.

hiwelo:
So, after realising that `jsdom` was already part of this repository with the version `12.2.0`, I wanted to see if I was able to rollback my branch to this version.
Unfortunately, it is not working with this version, but it is only working with the version `15+`.

Every test passed, so I think the upgrade should not be an issue.
And because it was already part of the repository, we should not have too many issues related to the usage of this package in the test suite.

@gnapse Thanks a lot for your reviews. Is there anything else I should update?

hiwelo:
Ok, so finally I was able to remove the specific testing I created earlier.
Thanks to the `jsdom` update to a 15+ version, I was able to `checkValidaty()` on a rendered HTMLElement using the existing helper. 🎉

Way cleaner this way @gnapse 🙂

hiwelo:
Hmm… The test is completed without any issue on my local machine… :/ 
Any idea why it's failing specifically on Travis?

gnapse:
> Hmm… The test is completed without any issue on my local machine… :/
> Any idea why it's failing specifically on Travis?

It seems the jsdom issue is still biting us

image

I don't mind if you need to revert this to how it was for the time being.

hiwelo:
Cool! I will revert my last commit and push it again.

hiwelo:
@gnapse I just rebased the branch to remove the last commit :+1: 
(gosh, I love git rebase so much haha)

Issue #115, opened 2019-06-07
Scoped package?


kentcdodds:
No pressure. I'm just wondering whether you've thought about moving to the scoped package format like the rest of the testing-library family of tools.

gnapse:
Yes, haven't had the time. Will look into it.

kentcdodds:
Ping. Would you like me to do this?

gnapse:
Yes, please. Thanks @kentcdodds. I'm not too confident on this packaging and publishing things, and was worried that I may break it if done in a hurry.

kentcdodds:
Alright! This is done 💯

PR #118, opened 2019-07-08
chore: update all packages and get 100% coverage

kentcdodds:
**What**:

This updates all the packages and gets us to 100% code coverage

**Why**:

Just routine maintenance as I prepare this for #115 

**How**:

`yarn upgrade-interactive --latest` and manually fixing a few things

Note that before there was one jest config, and then the tests were run with `--env jsdom` for running the same tests in both DOM and Node environments. Now we're using projects so we can run all the tests in a single test run.

However this means trouble if you want to scope watch mode down to either node or DOM. So I added `jest-watch-select-projects` which allows you to hit the `P` key during watch mode and toggle on/off each of the projects so you can keep things scoped to the area where you're working.

**Checklist**:

* [ ] Documentation N/A
* [x] Tests
* [x] Ready to be merged

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,7 +1,10 @@
-const jestConfig = require('kcd-scripts/jest')
+const config = require('kcd-scripts/jest')
 
-module.exports = Object.assign(jestConfig, {
-  testEnvironment: 'jest-environment-node',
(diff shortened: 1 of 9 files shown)


gnapse:
Thanks!

Issue #117, opened 2019-06-18
<select> element value change is not recognized properly by toHaveFormValues


gnapse:
* `jest-dom` version: 3.5.0

### Relevant code or config:

```js
const { container } = render(
  <form>
    <select name="priority">
      <option value="low">Low</option>
      <option value="medium">Medium</option>
      <option value="high">High</option>
    </select>
  </form>
)
const form = container.querySelector('form')
const select = container.querySelector('select')
fireEvent.change(select, { target: { value: 'high' } })
expect(form).toHaveFormValues({ priority: 'high' }) // <- fails here
```

The above test should pass, but it does not. It fails to recognize the change in the selected value of the `select`.

### Problem description:

After some digging I'm pretty sure this has something to do with the cause of [this issue in dom-testing-library](https://github.com/testing-library/dom-testing-library/issues/115). This issue was fixed in [this PR](https://github.com/testing-library/dom-testing-library/issues/124), and this is mention in that PR description:

> Since `selectedOptions` is not a reactive property, JSDOM was not updating to match the selected option in a select Element. The selected attribute is reactive.

We're using `selectedOptions` ourselves to check the selected options in a `select`. See [the relevant code](https://github.com/testing-library/jest-dom/blob/fd747bf470433e72d34be78e169fb213e0284e8f/src/utils.js#L141-L150).

### Suggested solution:

Maybe following the same approach as testing-library/dom-testing-library#124 and switch to use the `select` attribute instead.

jeanchung:
Hi! I just made a PR to fix this issue. ^

weyert:
Thank you, you rock :D

PR #126, opened 2019-08-18
fix: recognize changed value of select

jeanchung:
**What**:
Fixes https://github.com/testing-library/jest-dom/issues/117, adds a test for this use case, and adds me as a contributor.

**Why**:
The `getSelectValue()` util was using `selectedOptions` to get the selected options for a `select`. Because `selectedOptions` is not a reactive property, this was causing a bug where `toHaveFormValues()` would not recognize changed `select` values.

**How**:
This PR updates `getSelectValue()` to use `options` rather than `selectedOptions`, then filter for the selected options, similar to the way this was fixed in [dom-testing-library](https://github.com/testing-library/dom-testing-library/pull/124) (as suggested in the issue).

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/to-have-form-values.js
+++ b/src/__tests__/to-have-form-values.js
@@ -229,6 +229,24 @@ describe('.toHaveFormValues', () => {
         expect(form).toHaveFormValues(expectedValues)
       }).not.toThrowError()
     })
+
+    it('matches change in selected value of select', () => {
+      const oldValue = ''
+      const newValue = 'design'
+
+      const {container} = render(`
+        <form>
+          ${renderSelectSingle('category', 'Category', categories, oldValue)}
+        </form>
+      `)
+
+      const form = container.querySelector('form')
+      const select = container.querySelector('select')
+      expect(form).toHaveFormValues({category: oldValue})
+
+      select.value = newValue
+      expect(form).toHaveFormValues({category: newValue})
+    })
   })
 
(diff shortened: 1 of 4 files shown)


weyert:
Thank you for your contribution 💃

src/__tests__/to-have-form-values.js
```diff
@@ -331,3 +349,5 @@ function renderSelectMultiple(name, label, options, value = []) {
 function renderList(items, mapper) {
   return items.map(mapper).join('')
 }
+
+/* eslint max-lines-per-function:0 */
```
weyert: Just wondering why this is added?

jeanchung: The linter was throwing this error on line 108 due to the length of tests for `toHaveFormValues()`:

Screen Shot 2019-08-18 at 2 56 59 PM

I noticed that [this test file](https://github.com/testing-library/jest-dom/blob/master/src/__tests__/to-have-value.js) handled this problem this way, so I did the same. Let me know if another solution is preferred here!

weyert: Ah alright, seems fair. Guess, we could split it up. I think @kentcdodds wrote an article about avoiding nesting: https://kentcdodds.com/blog/avoid-nesting-when-youre-testing

Issue #125, opened 2019-08-16
Extend toBeValid to support the <form/> element


jsphstls:
### Describe the feature you'd like:

`toBeValid` and `toBeInvalid` could support checks against a `<form/>` element. `<form />` also responds to `checkValidity()`.

### Suggested implementation:

Add `'form'` to this list:
https://github.com/testing-library/jest-dom/blob/09f7f041805b2a4bcf5ac5c1e8201ee10a69ab9b/src/to-be-invalid.js#L4

### Describe alternatives you've considered:

`expect(form.checkValidity()).toBe(false)`
`expect(form.checkValidity()).toBe(true)`

### Teachability, Documentation, Adoption, Migration Strategy:

The current output does not need to be changed:
```
expect(element).toBeValid()

Received element is not currently valid:
      <form />
```

The current documentation does not need to be changed:

| This allows you to check if the value of a form element is currently valid.

| An element is valid if it is not having an aria-invalid attribute or having false as a value and returning true when calling checkValidity().

The documentation could be amended to include an example that checks a `<form/>`.

gnapse:
@jsphstls you're right. Are you up to make a PR? I may take this if you prefer. Let me know.

jsphstls:
Thanks @gnapse, I can work on it next week if that's alright. Should be quick.

jsphstls:
Opened a pull request ^

brunoxd13:
@gnapse @jsphstls I believe this issue can be closed

PR #127, opened 2019-08-19
feat: toBeValid supports <form/> element

jsphstls:
**What**:

`toBeValid` and `toBeInvalid` currently support checks against `input` elements by calling `checkValidity()`. <form /> also responds to `checkValidity()`.

**Why**:

Discussed in this issue:
https://github.com/testing-library/jest-dom/issues/125

**How**:

Added `'form'` to the list of tags that can be checked by `toBeValid` and `toBeInvalid`.

**Checklist**:

- [ x] Documentation
- [x] Tests
- [ ] Updated Type Definitions "N/A"
- [x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,13 +260,14 @@ expect(
 toBeInvalid()
 ```
 
-This allows you to check if an form element is currently invalid.
+This allows you to check if a form element, or the entire `form`, is currently
+invalid.
 
-An element is invalid if it is having an
+An `input`, `select`, `textarea`, or `form` element is invalid if it has an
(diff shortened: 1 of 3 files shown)


gnapse:
@jsphstls this looks good. Thanks again!

I wanted to ask you to update the README a little bit to reflect that not only form elements, but the form itself can also be check for validity? At least mention in the README doc for `toBeValid` and `toBeInvalid` that an entire form can be checked. No need to add html examples of it, but do it if you feel like it.

jsphstls:
Let me know how that looks. I noticed that the README example closely matched the test cases, so I changed the invalid element example to be simpler to read.

Issue #123, opened 2019-08-15
Make it so the TypeScript definitions work automatically without config


kentcdodds:
If we follow the path we took in [React Testing Library](https://github.com/testing-library/react-testing-library/pull/437) then that'll mean that people wont have to do anything [fancy](https://github.com/testing-library/jest-dom/issues/45#issuecomment-459651274) to make the typings/autocomplete show up which would be rad.

Specifically:

1. Add `testing-library__jest-dom` to `DefinitelyTyped`
2. Add `@types/testing-library__jest-dom` to `@testing-library/jest-dom` `dependencies`

I just tested it locally and that appears to work. If I have it in my node_modules then it just works™️

Anyone up for this?

MichaelDeBoey:
Happy to help out on this one! 🙂

kentcdodds:
Fantastic! Go for it!

gnapse:
Wouldn't this make it possible that the type definitions lag behind the latest features in the library? I understand that part of the motivation in doing this in RTL and DTL was this:

> There are no core teammembers with the TS knowledge needed to maintain the types, and they are not well tested or integrated with the codebase.

I'm no expert in TS, but I feel qualified enough to maintain the small needs of jest-dom. It's mainly about making sure that PRs that introduce new features or change the external interface make the appropriate updates.

I'll also check if #45 is indeed still an issue. If that's so, then I'd understand the need for this more. Will post my findings here.

gnapse:
My main gripe with DefinitelyTyped is that the type definitions are usually lagging behind and users are unable to use new features until the type definitions get up to speed with the latest developments in the repo.

Also, last time I checked DefinitelyTyped was this huge repo with all the type definitions of everything under the sun, all in the same place. It's hard to search for existing issues of specific libraries, since it's all under the same list of issues. I appreciate its existence because most libs don't provide typings, understandably, but I always thought (and my experience as a consumer of libraries has confirmed) that if the library itself provides the type definitions (e.g. Formik) the experience is so much better. I've been off using TS actively for a few months now, so not sure if any of the reasons for these perceptions have changed. I'd love to know others' input on all this, and if any of this was taken into consideration when making the decision to go in this direction in RTL/DTL.

jgoz:
DefinitelyTyped does have a slight advantage when it comes to global typings, since they will be implicitly included for normal TS projects (those that don’t specify a `types` tsconfig option).

[...]

PR #129, opened 2019-09-16
Update README.md

pauldariye:
Update installation instruction for Typescript

**What**:
- Documentation

**Why**:
- Update instruction for Typescript

**How**:

**Checklist**:

- [x] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ import {toBeInTheDocument, toHaveClass} from '@testing-library/jest-dom'
 expect.extend({toBeInTheDocument, toHaveClass})
 ```
 
-**Note** If you're using Typescript, make sure your setup file is a `.ts` and not a `.js` to include the necessary types.
+> Note: If you're using TypeScript, make sure your setup file is a `.ts` and not a `.js` to include the necessary types.
 
 
 ## Custom matchers
--- a/README.md
+++ b/README.md
@@ -93,6 +93,10 @@ Import `@testing-library/jest-dom/extend-expect` once (for instance in your
 [tests setup file]:
   https://jestjs.io/docs/en/configuration.html#setupfilesafterenv-array
 
+
+> Note: If you're using TypeScript, make sure your setup file is a `.ts` and not a `.js` to include the necessary types.
+
+
 ```javascript
 import '@testing-library/jest-dom/extend-expect'
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -106,9 +106,8 @@ import {toBeInTheDocument, toHaveClass} from '@testing-library/jest-dom'
 expect.extend({toBeInTheDocument, toHaveClass})
 ```
 
-> Note: when using TypeScript, this way of importing matchers won't provide the
-> necessary type definitions. More on this
-> [here](https://github.com/testing-library/jest-dom/pull/11#issuecomment-387817459).
+**Note** If you're using Typescript, make sure your setup file is a `.ts` and not a `.js` to include the necessary types.
```
gnapse: Nit pick:

```suggestion
**Note** If you're using TypeScript, make sure your setup file is a `.ts` and not a `.js` to include the necessary types.
```

gnapse: Also, an extra nit-pick: this kind of paragraphs that are meant as a side note in the README, and not part of the normal flow of reading, I'd rather have them be a note-like paragraph, preceding in each line with markdown notation `>`, just as the paragraph you replaced originally.

pauldariye:
> Thanks for your contribution. Indeed, there are things to be improved about this TypeScript aspect in the README. However, I'm not entirely sure this replacement is accurate with respect to the intentions of the text you're replacing.
> 
> The text being replaced was meant to convey the fact that picking and choosing to import just some of the custom matchers, and extent jest just with the ones you want, and not all, this won't make TypeScript happy. And using a setup file in TS as you mention won't fix that either.
> 
> Using a TS setup file will fix importing jest-dom in the usual manner. So I think the text you have included here, rather than replacing this part of the README, should be inserted, without replacing anything, before the paragraph in the README a few lines above this part, the one that starts with the sentence "Alternatively, you can selectively import only..."
> 
> Can you comment on these concerns of mine and clarify if I'm misreading something here?
> 
> (Also, as a side note, this PR is related to #123 so I'll mention that here to create the cross reference to that issue, which, when solved, may even change this update).

Hey @gnapse, I think you're right. I misread that. I'll put back that paragraph.

pauldariye:
I've updated per your comment and suggestion @gnapse and thanks!

PR #133, opened 2019-09-24
test: Remove test in node environment

eps1lon:
This may be a bit ignorant but are there any other DOM implementations
in node than jsdom? In any case this library needs a DOM so there's
no need to test in a node environment, right? 

The node environment was originally added to get 100% coverage but if this would reduce the coverage than it's either not actually test or not needed since you won't cover those branches when using this library.

I was working on a `inAccessibleMatcher` but using utils from `@testing-library/dom` does not work because they require a DOM

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -5,6 +5,5 @@ module.exports = {
   projects: [
     require.resolve('jest-watch-select-projects'),
     require.resolve('./tests/jest.config.dom'),
-    require.resolve('./tests/jest.config.node'),
   ],
 }
--- a/tests/jest.config.dom.js
+++ b/tests/jest.config.dom.js
@@ -4,6 +4,6 @@ const config = require('kcd-scripts/jest')
 module.exports = {
   rootDir: path.resolve(__dirname, '..'),
   displayName: 'jsdom',
-  testEnvironment: 'dom',
   ...config,
+  testEnvironment: 'jsdom',
 }
--- a/tests/jest.config.node.js
+++ b/tests/jest.config.node.js
@@ -1,9 +1,0 @@
-const path = require('path')
-const config = require('kcd-scripts/jest')
-
-module.exports = {
-  rootDir: path.resolve(__dirname, '..'),
-  displayName: 'node',
-  testEnvironment: 'node',
-  ...config,
-}


kentcdodds:
I believe the desire is that these matchers can be added when the DOM isn't globally accessible. There are people who use this and other Testing Library tools without a globally accessible DOM (they use jsdom directly).

eps1lon:
Sure you can do this. But in the actual environment the DOM will always be available globally as far as I know.

This is also a bit of a consistency problem. `lesting-library/dom` is not usable without a globally available DOM. `jest-dom` is. Not sure what makes `jest-dom` fundamentally different.

I would prefer requiring a globally available DOM. But if this is a strict no then we should at least make it consistent with `listing-library/dom`.

kentcdodds:
Actually, `@testing-library/dom` *does* work without the DOM globally available: https://github.com/testing-library/dom-testing-library/blob/fa662242e893d6cd35fd08deca3af1d600750657/tests/jest.config.node.js

This was a hard requirement for a few users, and it's not that difficult to support with a little care, so I don't mind requiring it.

eps1lon:
> Actually, @testing-library/dom does work without the DOM globally available: testing-library/dom-testing-library:tests/jest.config.node.js@fa66224

I need to double check but this was the whole reason I opened it: It does not: 

> I was working on a inAccessibleMatcher but using utils from @testing-library/dom does not work because they require a DOM

kentcdodds:
Ah, I see. Well, maybe we should discuss this with other maintainers 🤔

alexkrolick:
Having DTL importable without a global DOM was useful for processing the lib within Node and then doing something like injecting it into a Puppeteer context

eps1lon:
Yeah `@testing-library/dom` has separate tests for node in https://github.com/testing-library/dom-testing-library/blob/792c5b7a0037499784c8d95a4a78c4079f928acd/src/__node_tests__/

Currently it would fail in https://github.com/testing-library/dom-testing-library/blob/7f8de8943bdb46c2d855fb6f721c47351a942547/src/role-helpers.js#L18

> Having DTL importable without a global DOM was useful for processing the lib within Node and then doing something like injecting it into a Puppeteer context

Being importable and being usable are two different things though. Currently `jest-dom` is expected to be usable without a browser while `@testing-library/dom` is not. At least if you compare their test setups.

kentcdodds:
Ah, that was my bad. We shouldn't have merged that code as written. We should've instead used: `element.ownerDocument.defaultView.getComputedStyle(element)`.

eps1lon:
Sure but could clear up what the intention of this is? Should the library be importable without a global DOM or usable? Because these `defaultView` usages will litter the codebase and we need to change the testing setup in `@testing-library/dom` if we want to pursue this.

kentcdodds:
Actually I don't think there are many places `window` or`document` are used, so we're probably ok.

eps1lon:
Ok so we do need to update the testing-library/dom. I'll start work on it.

PR #135, opened 2019-10-02
Add types to package.json

### LINKED ISSUE #123: Make it so the TypeScript definitions work automatically without config

mrmckeb:
**What**:

Resolves #123.

**Why**:

This property in `package.json` tells TypeScript where to find the types for this `jest-dom`.

**How**:

A `lib` property was add to `package.json`.

A user must only import the below into any `.ts` or `.d.ts` file.
```ts
import '@testing-library/jest-dom/extend-expect';
```

**Checklist**:

- [ ] Documentation
- [x] Tests N/A
- [x] Updated Type Definitions N/A
- [x] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "0.0.0-semantically-released",
   "description": "Custom jest matchers to test the state of the DOM",
   "main": "dist/index.js",
+  "types": ".",
   "engines": {
     "node": ">=8",
     "npm": ">=6"


package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.0.0-semantically-released",
   "description": "Custom jest matchers to test the state of the DOM",
   "main": "dist/index.js",
+  "types": "extend-expect.d.ts",
```
jgoz: The problem with this is is that the `main` file, `dist/index.js`, does not extend the global `expect` object. So only doing `import @testing-library/jest-dom` will provide typings for the extensions but it will fail at runtime because `expect` was not extended.

mrmckeb: OK, sure. So we can instead use a folder. Let me test that.

mrmckeb: @jgoz This has been updated, and is still working for me locally.

package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.0.0-semantically-released",
   "description": "Custom jest matchers to test the state of the DOM",
   "main": "dist/index.js",
+  "types": ".",
```
jgoz: I don't think this actually does anything... `import '@testing-library/jest-dom/extend-expect';` from a TS file has always worked.

Again, the point of #123 was to cheat (a bit) and make the `extend-expect` typings available without needing any imports (i.e., make them "global" typings). But the only way to make them global typings is by moving them to DefinitelyTyped and defining them as global typings, rather than module typings.

This is something that the DT maintainers disagreed with, so barring some kind of TS compiler support for automatic global type resolution from types provided by npm packages, we are stuck.

mrmckeb: For me, this did make a difference. Without this, it didn't work - I checked a few times.

For me, this only worked otherwise if I did:
```ts
import '@testing-library/jest-dom/extend-expect.d'
```

This may be environmental.

gnapse:
What's the status of this? And what's its relation with #123? Would it solve that issue?

jgoz:
This PR does not solve #123, unfortunately. There is no way to solve #123 entirely within our project. The only way to do that involves the `@types/` hack that the DefinitelyTyped team did not accept (as far as I'm aware).

gnapse:
Ok, we'll continue the discussion about #123 over there.

Is there any value in merging this regardless of it solving #123?

jgoz:
The changes in this PR won't have an effect on library users, so I don't see a reason to merge. TypeScript can find the `index.d.ts` types without a `"types"` property in `package.json` and people will still need to manually include `extend-expect.d.ts` somewhere.

gnapse:
Closing this based on the previous comment, and also under the light of our most recent effort to solve #123 (see #175, #182 and DefinitelyTyped/DefinitelyTyped#37792).

PR #138, opened 2019-10-05
Fix .toHaveTextContent("")  causing false positive

### LINKED ISSUE #104: toHaveTextContent("") matches element with non-empty content

kandros:
**What**:

fix #104 

**Why**:

It can cause false-positive tests passing

**How**:

throwing when matching against an empty string, suggesting to use `.toBeEmpty()` instead

**Checklist**:

- [ ] Documentation N/A
- [X] Tests
- [X] Ready to be merged


gnapse:
@all-contributors please add @kandros for bug, test

PR #141, opened 2019-10-08
feat: Add toBeChecked matcher

### LINKED ISSUE #107: Add toBeChecked for <input type="radio" /> and <input type="checkbox" />

I want this to work:

```javascript
const input = document.createElement('input')
input.setAttribute('type', 'checkbox')
input.setAttribute('checked', true)
expect(input).toBeChecked()
input.setAttribute('checked', false)
expect(input).not.toBeChecked()
```

```javascript
expect(input.checked).toBeTrue()
```

This would need to be added to the docs like the rest of the assertions.

connormeredith:
Closes #107 

**What**:
Added a `toBeChecked` matcher.

**Why**:
It was not implemented in https://github.com/testing-library/jest-dom/pull/90 but I thought it would be a nice addition as there currently isn't a matcher to check the value of a checkbox without it being part of a form.

**How**:
Implemented along the same lines as `toHaveValue` except it's only limited to `<input type="checkbox />`.

I wasn't too sure on the best way to handle `<input type="radio" />` as it can be "checked" and also have a "value", so I've left that for now suggesting to use `toHaveFormValue` instead - similar to how `toHaveValue` handles this situation.

**Checklist**:
- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -923,8 +923,8 @@ toHaveValue(value: string | string[] | number)
 This allows you to check whether the given form element has the specified value.
 It accepts `<input>`, `<select>` and `<textarea>` elements with the exception of
 of `<input type="checkbox">` and `<input type="radio">`, which can be
-meaningfully matched only using [`toHaveFormValue`](#tohaveformvalues).
-`<input type="checkbox>` can also be matched with ['toBeChecked'](#tobechecked).
+meaningfully matched only using [`toBeChecked`](#tobechecked) or
+[`toHaveFormValue`](#tohaveformvalues).
 
 For all other form elements, the value is matched using the same algorithm as in
 [`toHaveFormValue`](#tohaveformvalues) does.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -97,7 +98,8 @@ Import `@testing-library/jest-dom/extend-expect` once (for instance in your
 import '@testing-library/jest-dom/extend-expect'
 ```
 
-> Note: If you're using TypeScript, make sure your setup file is a `.ts` and not a `.js` to include the necessary types.
```
connormeredith: It looks like prettier reformatted this, I'm happy to undo if required.

README.md
```diff
@@ -920,6 +924,7 @@ This allows you to check whether the given form element has the specified value.
 It accepts `<input>`, `<select>` and `<textarea>` elements with the exception of
 of `<input type="checkbox">` and `<input type="radio">`, which can be
 meaningfully matched only using [`toHaveFormValue`](#tohaveformvalues).
+`<input type="checkbox>` can also be matched with ['toBeChecked'](#tobechecked).
```
gnapse: What's the point of the single quotes around the tobeChecked link? Maybe you meant to use backticks?

connormeredith:
I've added support for `type="radio"` and `role="checkbox|radio"`. My knowledge of aria-checked is vague so I'm happy to make further changes to make it compliant with the spec if it isn't already. I've also not included the "indeterminate" state as per @kentcdodds [comment](https://github.com/testing-library/jest-dom/issues/107#issuecomment-529069879) on the original issue, I can always add it at a later date if it's needed 🙌

connormeredith:
@gnapse I've updated with `master`, is there anything else you need me to do to get this merged?

gnapse:
@all-contributors please add @connormeredith for code, test, doc

gnapse:
@all-contributors please add @connormeredith for code, test, doc

michaelmior:
Given that the goal for extra matchers is readability, I would expect the function to be called `toBeSelected` for radio buttons, as I don't ever see radio buttons referred to as being "checked" but "selected."

gnapse:
> Given that the goal for extra matchers is readability, I would expect the function to be called toBeSelected for radio buttons, as I don't ever see radio buttons referred to as being "checked" but "selected."

@michaelmior while I get your point, I'm not sure I agree. Both [the api at the native radio button level](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio) and the `aria-` side of things seem to agree that it is "checked" and not "selected". On the `aria-` side of things [`aria-checked`](https://www.digitala11y.com/aria-checked-state/) is supported in elements with `role="radio"`, while [`aria-selected`](https://www.digitala11y.com/aria-selected-state/) is used for different purposes and it's not even documented to make sense in a `role="radio"` element.

What do you think based on all this?

michaelmior:
That's a fair point. I guess I retract my statement in light of that :)

Belco90:
Just realized about this matcher was included. It's amazing, thank you!

PR #146, opened 2019-10-24
Add support for aria-disabled in toBeDisabled/toBeEnabled

### LINKED ISSUE #144: add aria-disabled in the `toBeDisabled` matcher

i have custom buttons (SVG's with click handlers) which i want to ensure are disabled under certain circumstances. Since these elements have disabled styling I've been using the `aria-disabled` attrubtue for this purpose (two birds one stone)

so i would like to be able to test as 
`expect(myCustomSvgButton).toBeDisabled()`

obiviously this is doable just now using `.toHaveAttribute` but i feel like this reduces the clarity of what's being checked and focuses on implementation details making the test brittle.

on a side note, maybe encouraging the use of `aria-*` attrs for communication of more complex state in the DOM is not a bad thing?

### LINKED ISSUE #145: latest version of jsdom (15.2.0) causes focus test to fail
- `@testing-library/jest-dom` version: 4.1.2
- `node` version: 12.6.0
- `npm` (or `yarn`) version: 6.11.3

### Relevant code or config:

...

### What you did:

```bash
npm run setup
```

### What happened:

```bash
[test] 
 FAIL  src/__tests__/to-have-focus.js
  ● .toHaveFocus
[test] 
    expect(element).toHaveFocus()
[test] 
    Expected
      <input id="focused" type="text" />
[test]     Received:
      null
[test] 
      14 |   focused.focus()
      15 | 
    > 16 |   expect(focused).toHaveFocus()
         |                   ^
      17 |   expect(notFocused).not.toHaveFocus()
      18 | 
      19 |   expect(() => expect(focused).not.toHaveFocus()).toThrowError()

      at Object.toHaveFocus (src/__tests__/to-have-focus.js:16:19)
[test] 
[test] 
[test] Test Suites: 2 failed, 32 passed, 34 total
[test] Tests:       2 failed, 126 passed, 128 total
[test] Snapshots:   8 passed, 8 total
```

pwolaq:
**What**:

* Add support for `aria-disabled` in `toBeDisabled/toBeEnabled` #144 
* Temporarily remove floating `jsdom` dependency until it is fixed #145 
* Few changes in README after running `npm run format`

Closes #144 
 
**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/to-be-disabled.js
+++ b/src/to-be-disabled.js
@@ -41,6 +41,10 @@ function isElementDisabled(element) {
   return FORM_TAGS.includes(getTag(element)) && element.hasAttribute('disabled')
 }
 
+function hasAriaDisabled(element) {
+  return element.getAttribute('aria-disabled') === 'true'
+}
+
 function isAncestorDisabled(element) {
   const parent = element.parentElement
   return (
@@ -52,7 +56,10 @@ function isAncestorDisabled(element) {
 export function toBeDisabled(element) {
   checkHtmlElement(element, toBeDisabled, this)
 
-  const isDisabled = isElementDisabled(element) || isAncestorDisabled(element)
+  const isDisabled =
+    hasAriaDisabled(element) ||
+    isElementDisabled(element) ||
+    isAncestorDisabled(element)
 
   return {
     pass: isDisabled,
@@ -71,7 +78,11 @@ export function toBeDisabled(element) {
 export function toBeEnabled(element) {
   checkHtmlElement(element, toBeEnabled, this)
 
-  const isEnabled = !(isElementDisabled(element) || isAncestorDisabled(element))
+  const isEnabled = !(
+    hasAriaDisabled(element) ||
+    isElementDisabled(element) ||
+    isAncestorDisabled(element)
+  )
 
   return {
     pass: isEnabled,
--- a/src/to-be-disabled.js
+++ b/src/to-be-disabled.js
@@ -41,7 +41,7 @@ function isElementDisabled(element) {
   return FORM_TAGS.includes(getTag(element)) && element.hasAttribute('disabled')
 }
 
-function hasAriaDisabled(element) {
+function isAriaDisabled(element) {
   return element.getAttribute('aria-disabled') === 'true'
 }
 
@@ -57,7 +57,7 @@ export function toBeDisabled(element) {
   checkHtmlElement(element, toBeDisabled, this)
 
   const isDisabled =
-    hasAriaDisabled(element) ||
+    isAriaDisabled(element) ||
     isElementDisabled(element) ||
     isAncestorDisabled(element)
 
@@ -79,7 +79,7 @@ export function toBeEnabled(element) {
   checkHtmlElement(element, toBeEnabled, this)
 
   const isEnabled = !(
-    hasAriaDisabled(element) ||
+    isAriaDisabled(element) ||
     isElementDisabled(element) ||
     isAncestorDisabled(element)
   )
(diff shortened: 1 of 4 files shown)


src/to-be-disabled.js
```diff
@@ -41,6 +41,10 @@ function isElementDisabled(element) {
   return FORM_TAGS.includes(getTag(element)) && element.hasAttribute('disabled')
 }
 
+function hasAriaDisabled(element) {
```
gnapse: nit pick: I'd rather call this `isAriaDisabled`.

src/to-be-disabled.js
```diff
@@ -52,7 +56,10 @@ function isAncestorDisabled(element) {
 export function toBeDisabled(element) {
   checkHtmlElement(element, toBeDisabled, this)
 
-  const isDisabled = isElementDisabled(element) || isAncestorDisabled(element)
+  const isDisabled =
+    hasAriaDisabled(element) ||
+    isElementDisabled(element) ||
```
gnapse: I wonder if there's some issue when having conflicting signals in an element that has different values for `disabled` and `aria-disabled`. Not saying we need to do anything about it, just wondering.

src/to-be-disabled.js
pwolaq: updated :)

gnapse:
Not sure if this is a `fix`, a `feat` or even a `BREAKING CHANGE` (maybe improbable but I can see how something that before was enabled can now make a test fail because it was really disabled via a `aria-disabled` attribute present 🤔, but not sure if I'm overthinking this) cc. @testing-library/core-maintainers

eps1lon:
The problem is that aria-disabled is only an atrribute for assistive technology while the disabled property has more implications (cant click or focus etc). So this matcher asserts very different things depending on which attribute is used. So I do think your test should reflect that.

gnapse:
Hmmmm, you bring up a good point @eps1lon but so much so that I now wonder if these two slightly different concepts of disabling an element should be expressed under the same matcher. I was indeed also worried about what to do when we have conflicting signals (e.g. `<span disabled aria-disabled="false" />`).

I'd say that we make it so we issue an error when we see that the `disabled` attribute does not agree with the `aria-disabled` attribute when both are present. But other than that, I'd keep the current logic of considering either one of them a valid signal of wether the target element is disabled or not. And not bother with wether it's a "hard" disable state (`disabled`) or a "soft" one (`aria-disabled`). Unless you can expand more on your concerns about this.

BTW, I'm also not sure how our tests should reflect what you bring up. Are you suggesting that we test that we can or cannot focus or click on elements accordingly?

eps1lon:
> BTW, I'm also not sure how our tests should reflect what you bring up. Are you suggesting that we test that we can or cannot focus or click on elements accordingly?

If someone researches how `disabled` should behave for each element then we can make this work. Otherwise I would caution against introducing this since it gives the impression that the `aria-disabled` __attribute__ asserts the same thing as the `disabled` __property__ (focus and click being the most important but other events as well like mouseEnter).

gnapse:
Just wanted to note that the discussion has apparently been moved to the issue (#144). Just so that we avoid having duplicate discussions about this, I guess the issue is better suited for this, since the objections are not with the implementation but with the feature proposal. Also @eps1lon already took the lead.

gnapse:
Hi,

I'm closing this for the same reasons stated in https://github.com/testing-library/jest-dom/issues/144#issuecomment-577235097. Feel free to continue the discussion there if you think there's something else yet to be said or discussed.

PR #150, opened 2019-10-26
Update type for Matchers to match @types/jest

bopfer:
**What**:

`@types/jest` recently had some updates to correct the type for `Matchers`.  This update is a breaking change for any TypeScript project that uses expect.extend.

See: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39243

**Why**:

`@testing-library/jest-dom` defines the `Matchers` type in `extend-expect.d.ts`.

**How**:

The change was simply to update the type to match `@types/jest`.

**Checklist**:

- [ ] Documentation - N/A
- [ ] Tests - N/A
- [x] Ready to be merged

### DIFF
--- a/extend-expect.d.ts
+++ b/extend-expect.d.ts
@@ -1,5 +1,5 @@
 declare namespace jest {
-  interface Matchers<R> {
+  interface Matchers<R, T> {
     /**
      * @deprecated
      */


jgoz:
@gnapse Any thoughts on this? Seems like it's blocking a lot of people.

gnapse:
I'll be merging it now as soon as the CI build passes. One question: does this closes #150? That issue description is not clear enough, but I see that it was attempted to be fixed in #153 and then that one got closed with a mention that it was a duplicate of this. So I just want to make sure we can close #150 after we merge this?

Issue #148, opened 2019-10-24
toHaveStyle failing for upper cased style rules in version 4.1.2


ppiyush13:
@testing-library/jest-dom library version: `4.1.2`

Consider the following div element:
```
<div id="simple" style="align-items:center">Simple Div</div>
```
I am using toHaveStyle to match CSS style rules as shown below

```
const div = document.querySelector("#simple");

expect(div).toHaveStyle('Align-items:center;');
```

**Expected:**
It should not throw error and proceed ahead

**Actual:**
It fails.

But below works fine
```
const div = document.querySelector("#simple");

expect(div).toHaveStyle('align-items:center;');
```

Seems that toHaveStyle needs CSS rules to be in lower case.
But this was **working as expected** in version `4.1.1`

ruhaise:
@ppiyush13  could you please try to put a codesandbox to reproduce this issue, im having some similar issue and i just wanted to compare :)

ppiyush13:
@ruhaise 
here is the codesandbox link https://codesandbox.io/s/jest-dom-issue-148-9lzc6

connormeredith:
@ppiyush13 can you let me know if https://github.com/testing-library/jest-dom/pull/154 fixes your issue?

ppiyush13:
@connormeredith I looked at the pull request https://github.com/testing-library/jest-dom/pull/154/files     . I think it should fix this issue

gnapse:
I'll be approving the PR because I confirmed css selectors are not case sensitive (to my surprise). I just wonder why are you specifying them in your tests with capitalized names? 🤔

ppiyush13:
@gnapse no specific reason for specifying CSS rules in upper case. I could have went ahead by making it lower case. I opened this issue because it was working fine till version  4.1.1

PR #154, opened 2019-10-29
Fixed issue where `.toHaveStyle` wouldn't match when the case was different.

connormeredith:
**What**:
Fixed issue where `.toHaveStyle` wouldn't match when the case was different: #148.

**Why**:
Appears to be a regression in https://github.com/testing-library/jest-dom/pull/81.

**How**:
Lowercased both the CSS property and value before comparing.

**Checklist**:
- [ ] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/to-have-style.js
+++ b/src/to-have-style.js
@@ -18,7 +18,9 @@ function getStyleDeclaration(document, css) {
 
 function isSubset(styles, computedStyle) {
   return Object.entries(styles).every(
-    ([prop, value]) => computedStyle.getPropertyValue(prop) === value,
+    ([prop, value]) =>
+      computedStyle.getPropertyValue(prop.toLowerCase()) ===
+      value.toLowerCase(),
   )
 }
 
--- a/src/__tests__/to-have-style.js
+++ b/src/__tests__/to-have-style.js
@@ -12,6 +12,7 @@ describe('.toHaveStyle', () => {
     const style = document.createElement('style')
     style.innerHTML = `
           .label {
+            align-items: center;
             background-color: black;
             color: white;
             float: left;
@@ -44,6 +45,10 @@ describe('.toHaveStyle', () => {
           color: white;
           font-weight: bold;
         `)
+
+    expect(container.querySelector('.label')).toHaveStyle(`
+        Align-items: center;
+      `)
   })
 
   test('handles negative test cases', () => {


src/to-have-style.js
```diff
@@ -18,7 +18,9 @@ function getStyleDeclaration(document, css) {
 
 function isSubset(styles, computedStyle) {
   return Object.entries(styles).every(
-    ([prop, value]) => computedStyle.getPropertyValue(prop) === value,
+    ([prop, value]) =>
+      computedStyle.getPropertyValue(prop.toLowerCase()) ===
+      value.toLowerCase(),
```
rkusa: I just came across this change with some failing tests after upgrading `@testing-library/jest-dom`. Not a big deal, though I'd like to point out that the value on the right hand side is converted to lowercase, while the one on the left hand side is not. So I guess it should be

```diff
    ([prop, value]) =>
-      computedStyle.getPropertyValue(prop.toLowerCase()) ===
+      computedStyle.getPropertyValue(prop.toLowerCase()).toLowerCase() ===
      value.toLowerCase(),
```

In my case it fails for comparing a `translateX`, so the comparison basically looks like `translateX(0px) vs translatex(0px)` now.

I would have suggest a PR, however, I am not sure if lowercasing the values is a good idea at all, since e.g. `font-family` values and urls are case-sensitive AFAIC. So the question probably is whether we don't lower-case the values at all, or add some exceptions?

gnapse: Maybe it's the `value` what should not be converted to lower case 🤔

connormeredith: Hmm this is tricky. I don't think CSS is case-sensitive: `transform: translateX(0px)` works the same as `transform: translatex(30px)`, `TRANSFORM: TRANSLATEX(30px)` and `transform: TRANSLATEx(30px)` (tested in chrome). I guess the best thing to do for now is to use your suggestion and lowercase the `prop`, `value` and `computedStyle.getPropertyValue`.

connormeredith: Yeah CSS is case insensitive: https://www.w3.org/TR/2011/REC-CSS2-20110607/syndata.html#characters
>The following rules always hold:
>
>- All CSS syntax is case-insensitive within the ASCII range (i.e., [a-z] and [A-Z] are equivalent), except for parts that are not under the control of CSS. For example, the case-sensitivity of values of the HTML attributes "id" and "class", of font names, and of URIs lies outside the scope of this specification. Note in particular that element names are case-insensitive in HTML, but case-sensitive in XML.
>...

So will need special cases for URIs and font names?

gnapse: So then what we need to do is what @rkusa suggests above. The only question is how to deal with font names and URIs (the only two things that can appear in css property values, because class names and IDs are only used in CSS selectors, which we do not process in `toHaveValues`).

dylang: Anything preventing us from returning `value.toLowerCase()` to just `value`?

gnapse: Nothing really. I can do it and will do it today. One thing though: I think it's better to add `.toLowerCase()` to the other operand in the comparison, rather than removing it from the `value`. @connormeredith @dylang @rkusa, what do you all think?

rkusa: Since we are talking about tests, the comparison of the values doesn't need to be case insensitive IMHO. So I'd either vote for lower-case with exceptions for fonts and URLs, or for not lowercasing the values at all. Since I can see how adding those exceptions might require quite some work to get it right, I'd be perfectly fine with the latter (not lowercasing the values).

gnapse: Good thinking. Let's avoid the full lowercasing, as it requires the special handling of the exceptions. Thanks for your input.

connormeredith: I agree, let's not lowercase the values 👍

gnapse: I very quickly created the PR for this: #163. Just did it fully online and in the browser just to kick it off. It probably need to adjustments in the tests, to which I'll get later today. But also to give some time to discuss there if things work.

BTW, to those of you affected, can you try out this very simple change by directly editing the corresponding file in your `node_modules` and run your tests again? Does it all work as expected when you introduce this change?

gnapse: Done in #164, thanks to @connormeredith. Should go out in a next release (should be 4.2.4, keep an eye for it being announced in that very PR in a while).

rkusa: Awesome, thanks @connormeredith  and @gnapse!

PR #161, opened 2019-11-09
Updated jest Matchers namespace declaration

shadrech:
**What**:

Jest namespace extension was out of date. The Matchers interface only accepts one generics variable

**Why**:

Though its stated as being deprecated I find once this is updated and I import as mentioned in [docs](https://github.com/testing-library/jest-dom#usage) typings work as expected

### DIFF
--- a/extend-expect.d.ts
+++ b/extend-expect.d.ts
@@ -1,5 +1,5 @@
 declare namespace jest {
-  interface Matchers<R> {
+  interface Matchers<R, T> {
     /**
      * @deprecated
      */
--- a/extend-expect.d.ts
+++ b/extend-expect.d.ts
@@ -1,5 +1,5 @@
 declare namespace jest {
-  interface Matchers<R, T> {
+  interface Matchers<R> {
     /**
      * @deprecated
      */
--- a/package.json
+++ b/package.json
@@ -50,6 +50,9 @@
     "jsdom": "^15.1.0",
     "kcd-scripts": "^1.4.0"
   },
+  "peerDependencies": {
+    "@types/jest": "^24.0.23"
+  },
   "eslintConfig": {
     "extends": "./node_modules/kcd-scripts/eslint.js",
     "rules": {


gnapse:
This seems to be at odds with #150. The change you're trying to introduce here is the exact opposite of that other PR. Can you read over that PR's description and the reasoning of why this was done? Because that one seem to suggest that what you want to introduce here is what's really out of date. Maybe we can involve the author of that PR in the discussion (@bopfer) as well as @jgoz who approved it.

bopfer:
#150 was done to sync up with `@types/jest`.  This comment explains the reasoning: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39243#issuecomment-550512994

jgoz:
Yeah, #150 introduced an implicit dependency upgrade to the latest Jest typings. I was worried this would happen (user upgrades to latest jest-dom, but not jest typings), but we don’t currently have a way to communicate this implicit dependency.

Maybe we could add `@types/jest` as an optional peer dep?

gnapse:
> user upgrades to latest jest-dom, but not jest typings)

Not sure what to do about this then, since either state of affairs affect some people.

> Maybe we could add @types/jest as an optional peer dep?

Would that definitely solve the issue? Are there any downsides?

jgoz:
> > user upgrades to latest jest-dom, but not jest typings)
> 
> Not sure what to do about this then, since either state of affairs affect some people.

I think it's simpler (for us) to always stay in sync with the latest Jest typings. My assumption is that people generally upgrade all jest-related packages together, including `jest-dom`.


> > Maybe we could add @types/jest as an optional peer dep?
> 
> Would that definitely solve the issue? Are there any downsides?

I wouldn't say it will "definitely solve" the issue since people can ignore peer dependency warnings, but it would at least provide some compatibility guidance for those using the typings.

gnapse:
Nice, seems reasonable.

@shadrech can you change this PR following what's discussed above?

shadrech:
Yh sure. Give me till the weekend and I'll work on it

shadrech:
Done @gnapse . Sorry for forced push. I had merged my local master branch which had some other changes I had made. Let me know if my changes suffice 👍

shadrech:
@jgoz Kindly review. Thanks

gnapse:
@jgoz how does this fare under the light of our recent work to fix #123? (context: #175, #182 and DefinitelyTyped/DefinitelyTyped#37792).

jgoz:
This can be closed. Once the types are published to DT, they will have `@types/jest` as a dependency.

Issue #165, opened 2019-11-18
asymmetricMatch for toHaveValue


benmonro:
### Describe the feature you'd like:

Much like https://github.com/testing-library/jest-dom/issues/91 it would be nice if `toHaveValue` also supported asymmetricMatch. for expect.stringContaining or expect.stringMatching etc.

### Suggested implementation:

use this.equal instead of lodash for comparison.

### Describe alternatives you've considered:

using toHaveAttribute("value", expect.stringContaining('foo')

### Teachability, Documentation, Adoption, Migration Strategy:

should be pretty straight forward to update all the things.

PR #166, opened 2019-11-19
Added jest matcher support for .toHaveValue()

connormeredith:
**What**:
Added jest matcher support for .toHaveValue().

**Why**:
Addressing issue on #165 and for greater flexibility in tests.

**How**:
Similar to https://github.com/testing-library/jest-dom/pull/93/.

**Checklist**:
- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/to-have-value.js
+++ b/src/to-have-value.js
@@ -1,11 +1,5 @@
 import {matcherHint} from 'jest-matcher-utils'
-import isEqualWith from 'lodash/isEqualWith'
-import {
-  checkHtmlElement,
-  compareArraysAsSet,
-  getMessage,
-  getSingleElementValue,
-} from './utils'
+import {checkHtmlElement, getMessage, getSingleElementValue} from './utils'
 
 export function toHaveValue(htmlElement, expectedValue) {
   checkHtmlElement(htmlElement, toHaveValue, this)
@@ -23,7 +17,7 @@ export function toHaveValue(htmlElement, expectedValue) {
   const expectsValue = expectedValue !== undefined
   return {
     pass: expectsValue
-      ? isEqualWith(receivedValue, expectedValue, compareArraysAsSet)
+      ? this.equals(receivedValue, expectedValue)
       : Boolean(receivedValue),
     message: () => {
       const to = this.isNot ? 'not to' : 'to'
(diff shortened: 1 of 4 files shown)


connormeredith:
@gnapse I'd like to add the same support to `toHaveFormValues` where you can provide a matcher per property or a matcher for the whole form, however I noticed there's a test that expects an array of values to match regardless of the order and I'm wondering how useful that is: https://github.com/testing-library/jest-dom/blob/master/src/__tests__/to-have-form-values.js#L132-L140.

**I guess this means that this PR would also remove that functionality from `toHaveValue`, since their implementation is very similar.**

As it stands we wouldn't be able to switch out the implementation for `this.equals` because it's checking the order of arrays under the hood and therefore wouldn't be able to support jest matchers for `.toHaveValue` and `.toHaveFormValue`. I'd personally be in favour of using `this.equals` and enforce array order for matchers and allow us to use jest matchers. What are your thoughts?

src/to-have-value.js
```diff
@@ -23,7 +17,7 @@ export function toHaveValue(htmlElement, expectedValue) {
   const expectsValue = expectedValue !== undefined
   return {
     pass: expectsValue
-      ? isEqualWith(receivedValue, expectedValue, compareArraysAsSet)
```
connormeredith: This change would mean we can no longer match arrays regardless of order, as per my [comment](https://github.com/testing-library/jest-dom/pull/166#issuecomment-555398045). But we wouldn't be able to support jest matchers without it, e.g. `expect.arrayContaining(['second', 'third'])`

gnapse:
@connormeredith sorry that this has slipped off our plate for so long. But it still sounds like an interesting path to follow. Do you have the bandwidth to resume working on this and ship it? If not that's ok. I got the gist of it and I think I can help finishing it too. Let me know and I'll plan accordingly.

tonivj5:
Any news on this? I have noticed that I can not use regexes to assert values with the `toHaveValue` matcher

gnapse:
Oh boy this is embarrassing. @connormeredith I'm ready to merge this. Can you take care of the merge conflict? That file needs to be removed, since types are no longer maintained by us.

PR #171, opened 2019-12-01
Fixes a bug where css.parse returns {} instead of a function

uzikilon:
**What**:

Replaces the way the CSS parse utility is imported

**Why**:

Fixes a bug where css.parse returns {} instead of a function in ts-jest
```
TypeError: (0 , _css.parse) is not a function
```

node : v13.2.0
"ts-jest: "^24.1.0",
"@testing-library/jest-dom": "^4.2.4",

**How**:

**Checklist**:

- [ ] Documentation "N/A"
- [ ] Tests "N/A"
- [ ] Updated Type Definitions "N/A"
- [x] Ready to be merged

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -7,7 +7,7 @@ import {
   printReceived,
   stringify,
 } from 'jest-matcher-utils'
-import {parse} from 'css'
+import parse from 'css/lib/parse'
 import isEqual from 'lodash/isEqual'
 
 class HtmlElementTypeError extends Error {


gnapse:
@uzikilon can you expand a little bit more on how/when/where this bug happens? I'm checking the [`css` lib documentation](https://github.com/reworkcss/css#readme) and I see that the recommended way to import it is from the module's default import and and not from a subpath. Maybe what we should do is import it like this instead:

```js
import css from 'css'

// Then use it like this:
css.parse(...)
```

uzikilon:
The bug is a specific typescript issue, accrues any time we run `toHaveStyle` on a typescript test loaded with ts-jest.

This is more of a workaround than a solution.
I suppose the real bug is in the CSS library and not jest-dom.

gnapse:
Oh thanks, that clarifies it.

Could you raise the issue in that library without closing this one, and you can reference this issue over there? I'd rather not introduce these kind of ad-hoc workarounds here, at least not until we check if there's a fix at the source.

PR #172, opened 2019-12-04
Add toBeSelected custom matcher

Kiarash-Z:
This new matcher checks whether an element has aria-selected of true or not.

**What**:

I added a new custom matcher called `toBeSelected`.

**Why**:

This is useful for checking state of selection for widgets which include single or multi selection like date pickers, form elements like select boxex or any `role=grid` item.

**How**:

By checking the value of `aria-selected`. There are two new files in `/src` and `/src/__tests__` corresponding to matcher name. `README.md` has been updated accordingly. Type Definitions are updated too. I added my name in `.all-contributorsrc` as well.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/extend-expect.d.ts
+++ b/extend-expect.d.ts
@@ -25,6 +25,6 @@ declare namespace jest {
     ): R
     toHaveValue(value?: string | string[] | number): R
     toBeChecked(): R
-    toBeEmpty(): R
+    toBeSelected(): R
   }
 }
--- a/extend-expect.d.ts
+++ b/extend-expect.d.ts
@@ -25,5 +25,6 @@ declare namespace jest {
     ): R
     toHaveValue(value?: string | string[] | number): R
     toBeChecked(): R
+    toBeEmpty(): R
   }
 }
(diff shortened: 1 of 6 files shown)


extend-expect.d.ts
```diff
@@ -25,5 +25,6 @@ declare namespace jest {
     ): R
     toHaveValue(value?: string | string[] | number): R
     toBeChecked(): R
+    toBeEmpty(): R
```
gnapse: Shouldn't this be `toBeSelected` instead?

```suggestion
    toBeSelected(): R
```

Kiarash-Z:
Thanks @gnapse, for the clarification 😊Sorry if I didn't open an issue first. 

> If we continue down this path, we may end up with other matchers such as toBeExpanded (aria-expanded) and toHavePopup (aria-haspopup).

Completely agreed! The fact is that I needed such matcher in my project because I had used that `aria-selected` many times. I thought that it would be nice if I add this to jest dom as well. However, it's rather simple and it can be implemented using a few lines of code.  I'm gonna close this PR

gnapse:
Thanks for being so understanding. And do not hesitate to come back here with other feautre requests or bug reports.

PR #182, opened 2020-01-16
Remove extend-expect typings

jgoz:
**What**:

BREAKING CHANGE: moving typings to `@types/testing-library__jest-dom` (see https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37792)

**Why**:

#123.

**How**:

- Deleted built-in TypeScript definitions
- Opened PR in DefinitelyTyped with new definitions
- Future PR: add dependency on `@types/testing-library__jest-dom` to this project

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Ready to be merged
      

After this PR, I think we can release version 5.0. Then the type definitions in the DefinitelyTyped PR will match the latest release. After that is merged, we can add the dependency and ship 5.0.1, finally closing the related issue.


jgoz:
Yeah, that’s right. Not ideal.

Instead of deleting the typings in this PR I could just have them apply to the index file instead of extend-expect. Then v5 would ship with typings. And 5.0.1 would swap the internal ones for the DT ones.

Thoughts on that?

JasonMore:
This just bit me. Can you guys update the documentation on the new setup process? @gnapse @jgoz Right now I don't see any types definitions using the current documentation, I had to search the commit log

```
"devDependencies": {
    "@testing-library/jest-dom": "5.0.0",
```

```
Error:(66, 27) TS2339: Property 'toHaveTextContent' does not exist on type 'JestMatchersShape<Matchers<void, HTMLElement>, Matchers<Promise<void>, HTMLElement>>'.
```

gnapse:
@JasonMore v5.0.1 is out now, with types support back. Sorry for the inconvenience. And let us know of any remaining or new issues with this update.

PR #183, opened 2020-01-16
Preparing for types being provided via DefinitelyTyped

gnapse:
**What**:

#175 and #182

**Why**:

#123 and DefinitelyTyped/DefinitelyTyped#37792

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged


README.md
```diff
@@ -110,7 +111,10 @@ Alternatively, you can selectively import only the matchers you intend to use,
 and extend jest's `expect` yourself:
 
 ```javascript
-import {toBeInTheDocument, toHaveClass} from '@testing-library/jest-dom'
+import {
+  toBeInTheDocument,
+  toHaveClass,
+} from '@testing-library/jest-dom/matchers'
```
jgoz: I just noticed - I don't think this import will actually work since there is no `matchers.js` at the solution root

gnapse: Oops, you're right. How do you think we should solve it?

jgoz: I guess the two options are 1) change the example to import from `@testing-library/jest-dom/src/matchers` or 2) add `matchers.js` to the root that does `export * from "./src/matchers"`...

gnapse: Will go for the second options. Will ping you later when it's done.

I will leave merging this to you, so that we can minimize the time 5.0.0 is the latest version. Let's merge it at a time when you know we can get the remaining bits done relatively quick after this.

jgoz: OK sounds good. Once I merge this, I will update the typings PR which should pass the build (since there will be a real 5.0 version). Hopefully it will get merged and released soon after that.

gnapse: @jgoz done d68ccd7.

PR #184, opened 2020-01-17
Fix toBeVisible ignoring Details element

lourenci:
**What**:

`toBeVisible` is ignoring an element inside a not opened `<details />`.

E.g. This div should be reported as a not visible element.
```jsx
<details>
  <summary>Title</summary>
  <div>This div is not visible</div>
</details>
```

[CodeSandbox](https://codesandbox.io/s/long-mountain-d92jm?fontsize=14&hidenavigation=1&module=%2Fsrc%2FApp.spec.js&previewwindow=tests&theme=dark)

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged
      

`<summary />` is visible whether the `<details />` is opened or not, this is not resolved in this PR yet.


src/__tests__/to-be-visible.js
```diff
@@ -32,4 +40,9 @@ test('.toBeVisible', () => {
   expect(() =>
     expect(container.querySelector('p')).toBeVisible(),
   ).toThrowError()
+
+  expect(queryByTestId('visible-details').querySelector('div')).toBeVisible()
+  // expect(queryByTestId('visible-details').querySelector('summary')).toBeVisible()
+  // expect(queryByTestId('hidden-details').querySelector('summary')).toBeVisible()
```
gnapse: Either use these or remove them.

src/to-be-visible.js
```diff
@@ -14,10 +14,16 @@ function isStyleVisible(element) {
   )
 }
 
+function hasElementAVisibleAttribute(element) {
```
gnapse: I'd recommend renaming this to `isDetailsElementOpen`.

src/to-be-visible.js
```diff
@@ -14,10 +14,16 @@ function isStyleVisible(element) {
...
+  return element.nodeName === 'DETAILS'
+    ? element.hasAttribute('open')
```
gnapse: I'm not sure this will work when the user manually toggles the details panel visibility. This is checking for the hard-coded presence of the attribute, and I think you need to check for the `element.open` property.

I'm basing my assumptions here on the MDN documentation. See [the "Events" section of the `details` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#Events).

eps1lon: I think this is good enough as a MVP. The underlying issue is that we check attributes instead of properties. For both `open` and `hidden` we should consult the DOM properties instead of HTML attributes.

src/to-be-visible.js
eps1lon: It's checking if the element is invisible due to certain HTML attributes such as `hidden` or `open`. It's actually incorrect since this would consider `<details hidden open>` as being visible.

I reconsidered landing this as-is. I think checking for `!(element.hidden || element.nodeName === 'DETAILS' && !element.open)` is the correct way to handle every combination

src/to-be-visible.js
gnapse: > The underlying issue is that we check attributes instead of properties. For both open and hidden we should consult the DOM properties instead of HTML attributes.

That's what I was suggesting. Without it, this selector could give an incorrect result in an actual web page where the details has been toggled open after the initial render.

src/to-be-visible.js
gnapse: I still prefer the more descriptive "isDetailsElementOpen". It describes what the function does, not how it does it.

gnapse: It also reflects the fact that this operates specifically on a `<details />` element.

lourenci:
> I'd like to see a test that does not add the `open` attribute manually to the `details` element, but that starts with a closed details element, then programmatically clicks on the summary, and then the matcher says it's visible.

Hi, @gnapse. I think this makes sense. How can we organize this specs? What do you think about something like that:

```js
describe('.toBeVisible')
  it('returns the visibility of an element')
  //the actual specs
  
  describe('with a <details /> element')
    describe('when the details is opened')
      it('returns true')

      describe('when the user clicks on the summary')
        it('returns false')

    describe('when the details is not opened')
      it('returns false')

      describe('when the user clicks on the summary')
        it('returns true')
```

src/__tests__/to-be-visible.js
lourenci: Sorry, it'll be removed. The PR is in wip yet.

src/to-be-visible.js
lourenci: Hi, I think it has the responsibility to operates in any element to check for an attribute about the element visibility, just like @eps1lon has mentioned. I've to check the `details` first because of it operates on an exception about its visibility attribute (it doesn't use the `hidden`). But, it's up to you...

> I think checking for `!(element.hidden || element.nodeName === 'DETAILS' && !element.open)` is the correct way to handle every combination

I'll make a spec to ensure that.

gnapse:
@lourenci certainly! I'm more than open to begin splitting more our tests to cover the different scenarios separately. So go ahead with what you propose.

src/to-be-visible.js
gnapse: Got it. Now I understand where my reasoning was flawed. Thanks to you two for clearing that up.

Now, I'll slightly insist in calling the function by what it does. I do not care if it checks for attributes. In my mind this function checks if the given element is considered visible. Isn't it?

lourenci: I've renamed to a name that matches the pattern applied to check the style visibility. Now we have something like that:
```js
isElementVisible => isStyleVisible && isAttributeVisible
```

lourenci:
I'm not sure about how we can handle the `summary` element, it's always visible except when some parent has a `[hidden]`. I can't think of anything better than that right now.

If you're ok with that, the PR is ok to be merged.

Thanks for the library.

gnapse:
> I'm not sure about how we can handle the summary element, it's always visible except when some parent has a [hidden]. I can't think of anything better than that right now.

I believe the `summary` element follows the rule of any other element. If it's `<summary hidden>` for example, it will be hidden, regardless of wether the parent `details` is open or not.

Will review and will let you know.

lourenci:
> Nested `details` elements

It's already ok. I've just created a context for that.

> Deeply nested elements inside the `details`

It's already ok. I've just created a context for that.

src/__tests__/to-be-visible.js
```diff
@@ -1,35 +1,247 @@
...
+          ).not.toBeVisible()
+        })
+
+        it('returns true to the nested details summary', () => {
+          expect(
+            subject.container.querySelector('details > details > summary'),
+          ).toBeVisible()
+        })
+
+        it('returns true to the nested details content', () => {
```
lourenci:
```suggestion
        it('returns true to the outer details content', () => {
```

src/__tests__/to-be-visible.js
```diff
@@ -1,35 +1,247 @@
...
+      it('returns false to the details content', () => {
+        expect(subject.container.querySelector('div')).not.toBeVisible()
+      })
+
+      it('returns false to the details summary', () => {
+        expect(subject.container.querySelector('summary')).not.toBeVisible()
+      })
+    })
+
+    describe('when there is a nested <details />', () => {
```
lourenci:
```suggestion
    describe('with a nested <details /> element', () => {
```

gnapse:
I'm deferring merging this a bit, in an attempt to make sure that our next release v5.0.1 is the one with fixed types support (see DefinitelyTyped/DefinitelyTyped#37792).

PR #187, opened 2020-01-17
Remove lines about using fireEvent to gain focus or blur

### LINKED ISSUE #186: Misleading documentation for toHaveFocus
- `@testing-library/jest-dom` version: 5.0.0
- `node` version: v13.3.0
- `npm` (or `yarn`) version: 6.13.2

### Relevant code or config:
The unit test of [`toHaveFocus`](https://github.com/testing-library/jest-dom/blob/master/src/__tests__/to-have-focus.js) extended
```js
import document from './helpers/document'

test('.toHaveFocus', () => {
  const {container} = render(`
      <div>
        <label for="focused">test</label>
        <input id="focused" type="text" />
        <button type="submit" id="not-focused">Not Focused</button>
      </div>`)

  const focused = container.querySelector('#focused')
  const notFocused = container.querySelector('#not-focused')

  document.body.appendChild(container)
  focused.focus()

  expect(focused).toHaveFocus()
  expect(notFocused).not.toHaveFocus()

  expect(() => expect(focused).not.toHaveFocus()).toThrowError()
  expect(() => expect(notFocused).toHaveFocus()).toThrowError()

  // The proceeding lines were added to match the example in documentation
  focused.blur()
  expect(focused).not.toHaveFocus()

  focused.dispatchEvent(
    new focused.ownerDocument.defaultView.FocusEvent('focus', {
      bubbles: false,
      cancelable: false,
    }),
  )
  expect(focused).toHaveFocus()
})
```

### What you did:

I was trying to follow the documentation of `toHaveFocus`: https://github.com/testing-library/jest-dom#tohavefocus

It also provides the following example:
> Using DOM Testing Library
```js
const input = queryByTestId(container, 'element-to-focus')

fireEvent.focus(input)
expect(input).toHaveFocus()

fireEvent.blur(input)
expect(input).not.toHaveFocus()
```

### What happened:

As a result I expected my input would report having focus after calling fireEvent.

### Reproduction:

The example I provided in the beginning, matches what `dom-testing-library` does when one calls `fireEvent.focus(input)`

When I run the modified tests, it fails with:

```
 FAIL   jsdom  src/__tests__/to-have-focus.js
  ● .toHaveFocus

    expect(element).toHaveFocus()

    Expected
      <input id="focused" type="text" />
    Received:
      <body><div>
          <div>
            <label for="focused">test</label>
            <input id="focused" type="text" />
            <button id="not-focused" type="submit">Not Focused</button>
          </div></div></body>

      31 |     }),
      32 |   )
    > 33 |   expect(focused).toHaveFocus()
         |                   ^
      34 | })
      35 | 

      at Object.<anonymous> (src/__tests__/to-have-focus.js:33:19)
```
Because just firing an event doesn't change the focused element in jsdom.

### Problem description:

The problem is that either documentation is misleading, or this needs to be fixed in the implementation of `dom-testing-library`.

### Suggested solution:

To avoid confusion, the example with markdown can be removed. Of course it would be best to fix this in `dom-testing-library` instead, by calling `input.focus()`, when one calls `fireEvent.focus(input)`

jzarzeckis:
Close #186 

**What**:

Removing lines in documentation that state you can use `fireEvent.focus(element)`

**Why**:

To avoid confusion. More details & example provided in #186 

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -732,18 +732,6 @@ input.blur()
 expect(input).not.toHaveFocus()
 ```
 
-##### Using DOM Testing Library
-
-```javascript
-const input = queryByTestId(container, 'element-to-focus')
-
-fireEvent.focus(input)
-expect(input).toHaveFocus()
-
-fireEvent.blur(input)
-expect(input).not.toHaveFocus()
-```
-
 <hr />
 
 ### `toHaveFormValues`


weyert:
Yes, or people would just use `user-event` instead ;)

gnapse:
Thanks for this contribution, and for noticing this in the first place.

Indeed, using `user-event` can fix this. But for the context of our documentation, I'd say I second @eps1lon's suggestion: instead of removing the entire example, we can change it so it calls `element.focus()` instead of `fireEvent.focus()`.

jzarzeckis:
@eps1lon if you look at the diff, you will see that the example with `element.focus()` hasn't been removed:

````diff
This allows you to assert whether an element has focus or not.
#### Examples
```html
<div><input type="text" data-testid="element-to-focus" /></div>
```
##### Using document.querySelector
```javascript
const input = document.querySelector(['data-testid="element-to-focus"'])
input.focus()
expect(input).toHaveFocus()
input.blur()
expect(input).not.toHaveFocus()
```

- ##### Using DOM Testing Library
-
- ```javascript
- const input = queryByTestId(container, 'element-to-focus')
-
- fireEvent.focus(input)
- expect(input).toHaveFocus()
-
- fireEvent.blur(input)
- expect(input).not.toHaveFocus()
- ```
````

gnapse:
@jzarzeckis indeed. But the existence of two examples was not because one used `element.focus()` and the other `fireEvent.focus()`. Both examples exist so that one shows an example that uses `@testing-library/dom` and the other shows how an example looks like when using native DOM APIs.

So your fix should be about making the example that was incorrectly using `fireEvent` to now use `element.focus()`.

If you wonder why we show duplicate examples like this, indeed, I share the pain, and I've been meaning to remove that duplication. I'd accept your PR as is, but this duplication is still needed to keep consistency. I'll remove that duplication in our documentation soon, but after your PR is merged nevertheless.

gnapse:
Follow up: created #188 to solve that documentation duplication for good. In case anyone wants to take a look or chime in.

gnapse:
@all-contributors please add @jzarzeckis for doc

PR #196, opened 2020-01-29
Enhance toHaveStyle to accept JS as css

### LINKED ISSUE #179: `toHaveStyle` support passing an object

Pass an object to the `toHaveStyle` matcher. The object shape would be the same as the style object in React.

This would allow the tests to match the code conventions of a project using object-based CSS-in-JS and offer type checking/code completion.

Offered as an alternative for the string function signature.

```javascript
expect(element).toHaveStyle({
  padding: '10px',
  marginTop: 5
});
```

lourenci:
This is a first naive implementation to resolve #179 .

I don't know the best way to generate a css text from an object, so I'm underlying to the DOM to do that.

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/to-have-style.js
+++ b/src/__tests__/to-have-style.js
@@ -164,6 +164,8 @@ describe('.toHaveStyle', () => {
       backgroundColor: 'red',
       height: '100%',
     })
-    // expect(container.querySelector('.label')).not.toHaveStyle({ whatever: 'anything' })
+    expect(container.querySelector('.label')).not.toHaveStyle({
+      whatever: 'anything',
+    })
   })
 })
--- a/src/__tests__/to-have-style.js
+++ b/src/__tests__/to-have-style.js
@@ -1,6 +1,7 @@
 import {render} from './helpers/test-utils'
 import document from './helpers/document'
 
+// eslint-disable-next-line max-lines-per-function
 describe('.toHaveStyle', () => {
   test('handles positive test cases', () => {
     const {container} = render(`
@@ -143,5 +144,26 @@ describe('.toHaveStyle', () => {
     expect(container.querySelector('.label')).not.toHaveStyle(
       'whatever: anything',
     )
+  })
+
+  test('handles styles as object', () => {
+    const {container} = render(`
+      <div class="label" style="background-color: blue; height: 100%">
+        Hello World
+      </div>
+    `)
+
+    expect(container.querySelector('.label')).toHaveStyle({
+      backgroundColor: 'blue',
+    })
+    expect(container.querySelector('.label')).toHaveStyle({
+      backgroundColor: 'blue',
+      height: '100%',
+    })
+    expect(container.querySelector('.label')).not.toHaveStyle({
+      backgroundColor: 'red',
+      height: '100%',
+    })
+    // expect(container.querySelector('.label')).not.toHaveStyle({ whatever: 'anything' })
   })
 })
--- a/src/to-have-style.js
+++ b/src/to-have-style.js
@@ -48,8 +48,19 @@ function expectedDiff(expected, computedStyles) {
   return diffOutput.replace(`${chalk.red('+ Received')}\n`, '')
 }
 
-export function toHaveStyle(htmlElement, css) {
+function getCss(document, css) {
+  if (typeof css === 'object') {
+    const sandboxElement = document.createElement('div')
+    Object.assign(sandboxElement.style, css)
+    return sandboxElement.style.cssText
+  }
+
+  return css
+}
+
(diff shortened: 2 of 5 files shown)


src/__tests__/to-have-style.js
```diff
@@ -1,6 +1,7 @@
 import {render} from './helpers/test-utils'
 import document from './helpers/document'
 
+// eslint-disable-next-line max-lines-per-function
```
gnapse: @kentcdodds I've been meaning to raise this issue since I started jest-do: what's the value of this eslint rule? Especially in tests it makes little sense. Asking you because I believe this comes from `kcd-scripts`, isn't it? I know we could go and disable it in our eslint overrides, but maybe this can be disabled at the root.

src/to-have-style.js
```diff
@@ -48,8 +48,19 @@ function expectedDiff(expected, computedStyles) {
   return diffOutput.replace(`${chalk.red('+ Received')}\n`, '')
 }
 
-export function toHaveStyle(htmlElement, css) {
+function getCss(document, css) {
+  if (typeof css === 'object') {
+    const sandboxElement = document.createElement('div')
+    Object.assign(sandboxElement.style, css)
+    return sandboxElement.style.cssText
+  }
```
gnapse: Wow, this was surprisingly simple. Looking forward to have this PR marked as ready.

src/__tests__/to-have-style.js
kentcdodds: Well, I rarely bump into violations of this rule because I don't use describe blocks. You can feel free to disable it in your overrides, but I think I want to keep it around in my config.

lourenci: What about removing just for specs files? Since we don't have a source analysis tool, this prevents long method bad smells.

lourenci:
I think it's ok to be reviewed and merged. I've never updated the types definitions on DefinitelyTyped, how can I do that? :)

src/utils.js
```diff
@@ -192,6 +192,12 @@ function compareArraysAsSet(a, b) {
   return undefined
 }
 
+function parseJStoCSS(document, css) {
+  const sandboxElement = document.createElement('div')
+  Object.assign(sandboxElement.style, css)
+  return sandboxElement.style.cssText
+}
```
kentcdodds: This is brilliant 👏

src/__tests__/to-have-style.js
lourenci: I'll open a PR for that.

gnapse:
> I've never updated the types definitions on DefinitelyTyped, how can I do that? :)

Good question. This was my fear when we switched to that instead of providing the types ourselves.

@jgoz @kentcdodds any experience with that? Any advice on how to proceed?

src/to-have-style.js
```diff
@@ -17,9 +17,12 @@ function getStyleDeclaration(document, css) {
 }
 
 function isSubset(styles, computedStyle) {
-  return Object.entries(styles).every(
-    ([prop, value]) =>
-      computedStyle.getPropertyValue(prop.toLowerCase()) === value,
+  return (
+    !!Object.keys(styles).length &&
```
gnapse: I guess you added this because we now need to return `false` if the object is empty, right?

Also, why not the more explicit alternative:

```suggestion
    Object.keys(styles).length > 0 &&
```

kentcdodds:
On the types, I pretty much just merge the PRs that make code changes whenever they're ready to be merged. Then eventually someone makes a PR to update DefinitelyTyped. Sometimes that happens before the merge, other times it happens after. In any case it happens eventually and it's fine 🤷‍♂️

jgoz:
> Good question. This was my fear when we switched to that instead of providing the types ourselves. Seems to go against co-location 😀
> 
> @jgoz @kentcdodds any experience with that? Any advice on how to proceed?

It's fairly straightforward. Fork the DT repo, make the change, open a PR. Their process is really streamlined for new PRs. The bot would likely ping me for a review since I'm the "author" of the typings and once I approve, a DT maintainer would merge it pretty quickly.

It can be a bit daunting because the DT repo is so huge, but it's just like any other npm package. Run `npm install`, `npm test` before you push, and you're good to go.

gnapse:
Ok, I'm merging this then. Still a bit unsettled that we lost co-location 🤷‍♂️

src/to-have-style.js
lourenci: > I guess you added this because we now need to return `false` if the object is empty, right?

Yes.

lourenci:
@jgoz I've just opened the PR there to update the types. https://github.com/DefinitelyTyped/DefinitelyTyped/pull/42022

PR #217, opened 2020-03-17
feat: add exact mode option for toHaveClass (#176)

SergiCL:
Added exact mode option for toHaveClass matcher, as discussed in #176

**How**
```html
<div data-test-id='test-element' class="hello world foo"></div>
```

```javascript
const testElement = getByTestId('test-element')

// toHaveClas is true becose has EXACTLY defined set of classes
expect(testElement).toHaveClass('hello', 'world', 'foo', {exact: true})

// toHaveClas is false because has te set of classes, but it also has 'foo'
expect(testElement).not.toHaveClass('hello', 'world', {exact: true})

// if exact option is not passed or is set as false, it will continue working as before.
expect(testElement).toHaveClass('hello', 'world')
expect(testElement).toHaveClass('hello', 'world', {exact: false})

```

If exact option is not passed or is false, it will continue working as before these changes.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/to-have-class.js
+++ b/src/to-have-class.js
@@ -10,7 +10,7 @@ function getExpectedClassNamesAndOptions(params) {
     options = lastParam
   } else {
     expectedClassNames = params.concat(lastParam)
-    options = {}
+    options = { exact: false }
   }
   return {expectedClassNames, options}
 }
--- a/src/to-have-class.js
+++ b/src/to-have-class.js
@@ -1,6 +1,20 @@
 import {matcherHint, printExpected} from 'jest-matcher-utils'
 import {checkHtmlElement, getMessage} from './utils'
 
+function getExpectedClassNamesAndOptions(params) {
+  const lastParam = params.pop()
+  let expectedClassNames, options
+
+  if (typeof lastParam === 'object') {
+    expectedClassNames = params
+    options = lastParam
+  } else {
+    expectedClassNames = params.concat(lastParam)
+    options = {}
+  }
+  return {expectedClassNames, options}
+}
+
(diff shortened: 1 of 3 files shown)


gnapse:
Oh no @SergiCL, I'm sorry it was not clear enough, but the API we decided when discussing this in #176 was to add a new matcher:

```js
expect(element).toHaveClassList('class1', 'class2', 'class3')
```

Could you change this PR to add that instead, and leave `toHaveClass` intact?

SergiCL:
Sure, no problem. I didn't understand it correctly.

By the way, do you know why Travis' validation is failing?
When I run it locally with the same version of node and npm it works without problem.

gnapse:
Can you bring up the naming change suggestion in the issue #176 (just so that the people that originally discuss this can chime in). I'll post my thoughts about it over there, but it kinda make sense your suggestion. Let's see what others think.

src/to-have-class.js
```diff
@@ -1,6 +1,20 @@
...
+function getExpectedClassNamesAndOptions(params) {
+  const lastParam = params.pop()
+  let expectedClassNames, options
+
+  if (typeof lastParam === 'object') {
+    expectedClassNames = params
+    options = lastParam
+  } else {
+    expectedClassNames = params.concat(lastParam)
+    options = {}
```
gnapse: Nit pick: should be `{ exact: false }`.

```suggestion
    options = { exact: false }
```

gnapse:
@all-contributors please add @SergiCL for code, test

Issue #221, opened 2020-03-24
`toHaveDisplayValue` custom matcher


cloud-walker:
I'm testing a labeled select that I query with `getByLabelText`, I then want to assert that the select has a default display name, but I can't find a jest matcher for that, here a snippet example:

```javascript
const select = screen.getByLabelText('Fruits')
expect(select).toBeVisible()
expect(select).toHaveDisplayValue('Select a fruit')
```

For now I think I should write a custom matcher myself, or a more complicated query 🤔

gnapse:
I'm sorry, not sure what you refer to with "default display name" of an element. Can you expand a bit?

cloud-walker:
Sure, I mean the currently selected item, so if I have:

```html
<label htmlFor="fruit">
  Fruits
</label>
<select name="fruit" id="fruit">
  <option value="" disabled selected>Select a fruit...</option>
  <option value="apple">Apple</option>
  <option value="ananas">Ananas</option>
  <option value="cherry">Cherry</option>
</select>
```

I want to be able to query the select by label text, and use the matcher to check that the displayed text is the one I want:

```javascript
const select = screen.getByLabelText('Fruits')
expect(select).toBeVisible()
expect(select).toHaveDisplayName('Select a fruit...')

// And if I want I can use it again to test changed values, like
fireEvent.change(select, {target: {value: {'apple'}}}})

expect(select).toHaveDisplayName('Apple') // with capital letter 😉 
```

cloud-walker:
There is already a testing-library query for that, but I want the matcher counterpart: https://testing-library.com/docs/dom-testing-library/api-queries#bydisplayvalue

cloud-walker:
Is more "display value" than "display name" sooorry!

gnapse:
Oh, I see. Makes sense. I'm all for it.

cloud-walker:
I'll see if I can make a PR

afontcu:
Solved in #223

PR #223, opened 2020-03-24
feat: Add `toHaveDisplayValue()` matcher

cloud-walker:
**What**:

Add `.toHaveDisplayValue()` custom matcher as discussed here: #221 

**Checklist**:

- [x] Documentation
- [x] Tests
- [ ] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -733,8 +733,8 @@ toHaveValue(value: string | string[] | number)
 
 This allows you to check whether the given form element has the specified value.
 It accepts `<input>`, `<select>` and `<textarea>` elements with the exception of
-of `<input type="checkbox">` and `<input type="radio">`, which can be
-meaningfully matched only using [`toBeChecked`](#tobechecked) or
+`<input type="checkbox">` and `<input type="radio">`, which can be meaningfully
+matched only using [`toBeChecked`](#tobechecked) or
 [`toHaveFormValues`](#tohaveformvalues).
 
 For all other form elements, the value is matched using the same algorithm as in
@@ -775,8 +775,12 @@ expect(selectInput).not.toHaveValue(['second', 'third'])
 toHaveDisplayValue(value: string | string[])
 ```
 
-This allows you to check whether the given `<select>` element has the specified
-value.
+This allows you to check whether the given form element has the specified
+displayed value (the one the end user will see). It accepts `<input>`,
+`<select>` and `<textarea>` elements with the exception of
+`<input type="checkbox">` and `<input type="radio">`, which can be meaningfully
+matched only using [`toBeChecked`](#tobechecked) or
+[`toHaveFormValues`](#tohaveformvalues).
 
 #### Examples
 
--- a/README.md
+++ b/README.md
@@ -46,7 +46,6 @@ clear to read and to maintain.
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 
-
 - [Installation](#installation)
 - [Usage](#usage)
 - [Custom matchers](#custom-matchers)
@@ -782,6 +781,12 @@ value.
 #### Examples
 
 ```html
+<label for="input-example">First name</label>
+<input type="text" id="input-example" value="Luca" />
+
+<label for="textarea-example">Description</label>
+<textarea id="textarea-example">An example description here.</textarea>
+
 <label for="single-select-example">Fruit</label>
 <select id="single-select-example">
   <option value="">Select a fruit...</option>
@@ -802,9 +807,13 @@ value.
 ##### Using DOM Testing Library
 
 ```javascript
+const input = screen.getByLabelText('First name')
+const textarea = screen.getByLabelText('Description')
 const inputSelect = screen.getByLabelText('Fruit')
 const inputMultiple = screen.getByLabelText('Fruits')
 
+expect(input).toHaveDisplayValue('Luca')
+expect(textarea).toHaveDisplayValue('An example description here.')
 expect(inputSelect).toHaveDisplayValue('Select a fruit...')
 expect(inputMultiple).toHaveDisplayValue(['Banana', 'Avocado'])
 ```
--- a/README.md
+++ b/README.md
@@ -814,8 +814,8 @@ const inputMultiple = screen.getByLabelText('Fruits')
 
 expect(input).toHaveDisplayValue('Luca')
 expect(textarea).toHaveDisplayValue('An example description here.')
-expect(inputSelect).toHaveDisplayValue('Select a fruit...')
-expect(inputMultiple).toHaveDisplayValue(['Banana', 'Avocado'])
+expect(selectSingle).toHaveDisplayValue('Select a fruit...')
+expect(selectMultiple).toHaveDisplayValue(['Banana', 'Avocado'])
 ```
 
 <hr />
--- a/README.md
+++ b/README.md
@@ -809,8 +809,8 @@ value.
 ```javascript
 const input = screen.getByLabelText('First name')
 const textarea = screen.getByLabelText('Description')
-const inputSelect = screen.getByLabelText('Fruit')
-const inputMultiple = screen.getByLabelText('Fruits')
+const selectSingle = screen.getByLabelText('Fruit')
+const selectMultiple = screen.getByLabelText('Fruits')
 
 expect(input).toHaveDisplayValue('Luca')
 expect(textarea).toHaveDisplayValue('An example description here.')
--- a/README.md
+++ b/README.md
@@ -778,13 +778,16 @@ value.
 #### Examples
 
 ```html
-<select data-testid="input-select">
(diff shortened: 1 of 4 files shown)


cloud-walker:
I've little experience with typescript, not sure if I should do something about type definitions 🤔

README.md
```diff
@@ -765,6 +766,44 @@ expect(selectInput).not.toHaveValue(['second', 'third'])
...
+  <option value="banana" selected>Banana</option>
+  <option value="ananas">Ananas</option>
+  <option value="avocado" selected>Avocado</option>
+</select>
+```
+
+##### Using DOM Testing Library
+
+```javascript
+const inputSelect = getByTestId('input-select')
```
eps1lon: We don't need the test-id here. Rather add an `aria-label` and use 
```suggestion
const inputSelect = getByRole('combobox', { name: 'what you wrote in aria-label' })
```

cloud-walker: Oh! I'm sorry! I copied the code from the tests T_T

I can use real labels if you prefer, like:

```html
<label for="single-select-example">Fruit</label>
<select id="single-select-example">
  <option value="">Select a fruit...</option>
  <option value="banana">Banana</option>
  <option value="ananas">Ananas</option>
  <option value="avocado">Avocado</option>
</select>

<label for="multiple-select-example">Fruits</label>
<select id="multiple-select-example">
  <option value="">Select a fruit...</option>
  <option value="banana" selected>Banana</option>
  <option value="ananas">Ananas</option>
  <option value="avocado" selected>Avocado</option>
</select>
```

```javascript
const inputSelect = getByLabelText('Fruit')
const inputMultiple = getByLabelText('Fruits')

expect(inputSelect).toHaveDisplayValue('Select a fruit...')
expect(inputMultiple).toHaveDisplayValue(['Banana', 'Avocado']
```

eps1lon: No need to apologize. This API is new. ByTestId is an escape hatch and we want to avoid it wherever possible.

alexkrolick:
This looks like it only supports `select` tags so far. Would you be open to making it work the same way as the byDisplayValue query?

> Returns the input, textarea, or select element that has the matching display value.

cloud-walker:
Hi, @alexkrolick I think it's a good idea, it could be a twin matcher with `toHaveValue` but for the display values. 

But I was wondering if do the extra work in another PR instead of adding more changes to this one 🤔

src/to-have-display-value.js
```diff
@@ -0,0 +1,34 @@
+import {matcherHint} from 'jest-matcher-utils'
+
+import {checkHtmlElement, getMessage} from './utils'
+
+export function toHaveDisplayValue(htmlElement, expectedValue) {
+  checkHtmlElement(htmlElement, toHaveDisplayValue, this)
+
+  if (htmlElement.tagName.toLowerCase() !== 'select') {
+    throw new Error(
+      '.toHaveDisplayValue() currently supports select elements only, try to use .toHaveValue() or toBeChecked() instead.',
```
tonivj5: Added dot to `toBeChecked`
```suggestion
      '.toHaveDisplayValue() currently supports select elements only, try to use .toHaveValue() or .toBeChecked() instead.',
```

src/__tests__/to-have-display-value.js
```diff
@@ -0,0 +1,62 @@
...
+
+  let errorMessage
+  try {
+    expect(queryByTestId('input')).toHaveDisplayValue('Banana')
+  } catch (err) {
+    errorMessage = err.message
+  }
+
+  expect(errorMessage).toMatchInlineSnapshot(
+    `".toHaveDisplayValue() currently supports select elements only, try to use .toHaveValue() or toBeChecked() instead."`,
```
tonivj5: Added dot to `toBeChecked` to be consistent :+1:

```suggestion
    `".toHaveDisplayValue() currently supports select elements only, try to use .toHaveValue() or .toBeChecked() instead."`,
```

gnapse:
> But I was wondering if do the extra work in another PR instead of adding more changes to this one

@cloud-walker I'd rather have it all at once, even if it takes longer. The reason being that the PR merge triggers a new release, and I'd rather have this feature released in its entirety.

This does not imply any pressure. Take your time, the same time you would've take to make the other PR.

Good work here overall, and thanks! Also thanks to @eps1lon for stepping in and reviewing.

cloud-walker:
Ok, input and textarea support added, I've excluded type=radio and type=checkbox, as the dont have a display value 🤔

cloud-walker:
Hi all, any news from this PR? 👀

README.md
```diff
@@ -768,6 +769,57 @@ expect(selectInput).not.toHaveValue(['second', 'third'])
...
 <hr />
 
+### `toHaveDisplayValue`
+
+```typescript
+toHaveDisplayValue(value: string | string[])
+```
+
+This allows you to check whether the given `<select>` element has the specified
+value.
```
gnapse, line 779: Seems this line here was not updated after you added support for other elements.

README.md
```diff
@@ -768,6 +769,57 @@ expect(selectInput).not.toHaveValue(['second', 'third'])
...
+</select>
+```
+
+##### Using DOM Testing Library
+
+```javascript
+const input = screen.getByLabelText('First name')
+const textarea = screen.getByLabelText('Description')
+const inputSelect = screen.getByLabelText('Fruit')
+const inputMultiple = screen.getByLabelText('Fruits')
```
gnapse, line 813: Nit pick, since these are not `input`.

```suggestion
const selectSingle = screen.getByLabelText('Fruit')
const selectMultiple = screen.getByLabelText('Fruits')
```

README.md
```diff
@@ -768,6 +769,57 @@ expect(selectInput).not.toHaveValue(['second', 'third'])
...
+
+expect(input).toHaveDisplayValue('Luca')
+expect(textarea).toHaveDisplayValue('An example description here.')
+expect(inputSelect).toHaveDisplayValue('Select a fruit...')
+expect(inputMultiple).toHaveDisplayValue(['Banana', 'Avocado'])
```
gnapse, line 818:
```suggestion
expect(selectSingle).toHaveDisplayValue('Select a fruit...')
expect(selectMultiple).toHaveDisplayValue(['Banana', 'Avocado'])
```

src/__tests__/to-have-display-value.js
```diff
@@ -0,0 +1,102 @@
...
+    <select id="fruits" data-testid="select">
+      <option value="">Select a fruit...</option>
+      <option value="ananas">Ananas</option>
+      <option value="banana">Banana</option>
+      <option value="avocado">Avocado</option>
+    </select>
+  `)
+
+  expect(queryByTestId('select')).toHaveDisplayValue('Select a fruit...')
+  expect(queryByTestId('select')).not.toHaveDisplayValue('Banana')
```
gnapse, line 14: Would be nice if the test also covered the case of changing the selected option programmatically to `banana` and then check that `toHaveDisplayValue('Banana')` pass.

And maybe do it for the other cases as well (textarea and input).

README.md
```diff
@@ -768,6 +769,57 @@ expect(selectInput).not.toHaveValue(['second', 'third'])
...
+<label for="single-select-example">Fruit</label>
+<select id="single-select-example">
+  <option value="">Select a fruit...</option>
+  <option value="banana">Banana</option>
+  <option value="ananas">Ananas</option>
+  <option value="avocado">Avocado</option>
+</select>
+
+<label for="mutiple-select-example">Fruits</label>
+<select id="multiple-select-example">
```
gnapse, line 799: Shouldn't it be `multiple`?

```suggestion
<select id="multiple-select-example" multiple>
```

src/to-have-display-value.js
```diff
@@ -0,0 +1,44 @@
...
+  const value =
+    tagName === 'select'
+      ? Array.from(htmlElement)
+          .filter(option => option.selected)
+          .map(option => option.textContent)
+          .toString()
+      : htmlElement.value
+
+  return {
+    pass: value === expectedValue.toString(),
```
gnapse, line 30: Hmmm, I'm not sure if comparing as string is the best approach. I now realize that this means that for `<select multiple />` you have to pass the array with the option display texts in the same order as they appear in the list.

cloud-walker: Yep, its for the `multiple` case, and at the moment I don't dislike too much the order enforcing thing, is indeed the same order that the user see when using those select 🤔

gnapse: Ok, let's leave it as is and we can improve upon it later if it proves to be a problem.

README.md
cloud-walker: Done, I've used the same copy of the `toHaveValue` matcher (and fixed a typo from it)

src/__tests__/to-have-display-value.js
cloud-walker: Done, I'm wondering why the repository is not using `fireEvent` from the testing-library core (or even userEvent) 🤔

gnapse:
@all-contributors add @cloud-walker for code, test, ideas

PR #228, opened 2020-04-08
feat: add switch role to toBeChecked

jonrimmer:
**What**:

Add support for the [ARIA switch role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Switch_role) to the `toBeChecked` matcher.

**Why**:

The `switch` role works the same as the `checkbox` role, with the switch's status exposed via `aria-checked`, but `toBeChecked()` cannot be used with a switch because only the `checkbox` and `radio` roles are currently implemented.

**How**:

Added `'switch'` to the list of allowed roles in `to-be-checked.js`

**Checklist**:

- [x] Documentation
- [x] Tests
- [ ] Updated Type Definitions — N/A (No type definitions in this project.)
- [x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,6 @@ clear to read and to maintain.
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 
-
 - [Installation](#installation)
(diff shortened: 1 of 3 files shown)


gnapse:
Thanks!

gnapse:
@all-contributors add @jonrimmer for code, test

PR #238, opened 2020-04-29
Fix documentation about adding to jest setup file

kraftman:
I tried to add this to jest using 'import '@testing-library/jest-dom' like the docs say, but it didn't work, 
what did work is 'import '@testing-library/jest-dom/extend-expect', so I've updated the docs to reflect it.

Apologies if I've missed something


gnapse:
Hmmm, it should work with how is currently documented. I'd rather fix that for you instead of making this change.

Can you provide a minimal example repo or codesandbox reproducing the issue?

kraftman:
I pulled out the minimal example I could from our project, and it seems to work fine. There must be some issue specific to our project, but I have no idea what. 

I guess I'll close this, sorry!

PR #242, opened 2020-05-06
Allow to use regex on toHaveDisplayValue

### LINKED ISSUE #239: `toHaveDisplayValue` doesn't support regex
`expect(userSelect).toHaveDisplayValue(/bezos/i)` doesn't work, instead it converts `/bezos/i` to string and tries to do a literal string match, as seen in the error:

```bash
    Expected element to have display value:
      /bezos/i
    Received:
      bezos
```

I'd like to stick with regex because the way we format HTML, it creates leading and trailing whitespaces, which prevent the match.

```html
	<option>
		{{u.name}}
	</option>
```

It won't be very pretty to have tests that look like `expect(userSelect).toHaveDisplayValue(" bezos ")`, ugly and error prone.

lourenci:
It closes #239.

All the current specs are kept. I just separate them in contexts to better describe the specs.

This implementation comes with one (good?) side effect: unordered multiple values didn't match before it.

```
<select id="fruits" data-testid="select" multiple>
  <option value="">Select a fruit...</option>
  <option value="ananas" selected>Ananas</option>
  <option value="banana">Banana</option>
  <option value="avocado" selected>Avocado</option>
</select>

// This was not true before this PR. It's true now.
expect(subject.queryByTestId('select')).toHaveDisplayValue([
  'Avocado',
  'Ananas',
])
```

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/to-have-display-value.js
+++ b/src/to-have-display-value.js
@@ -24,11 +24,13 @@ export function toHaveDisplayValue(htmlElement, expectedValue) {
       ? Array.from(htmlElement)
           .filter(option => option.selected)
           .map(option => option.textContent)
-          .toString()
-      : htmlElement.value
+      : [htmlElement.value]
+
+  const expectedValueArray = expectedValue instanceof Array ? expectedValue : [expectedValue]
+  const matchess = expectedValueArray.filter(expected => value.filter(valueFilter => matches(valueFilter, expected)).length).length
 
   return {
-    pass: matches(value, expectedValue),
+    pass: matchess === value.length && matchess === expectedValueArray.length,
     message: () =>
       getMessage(
         matcherHint(
--- a/src/to-have-display-value.js
+++ b/src/to-have-display-value.js
@@ -1,7 +1,5 @@
 import {matcherHint} from 'jest-matcher-utils'
-import {matches,checkHtmlElement, getMessage} from './utils'
-
-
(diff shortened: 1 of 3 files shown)


src/to-have-display-value.js
```diff
@@ -38,7 +40,25 @@ export function toHaveDisplayValue(htmlElement, expectedValue) {
...
+        .filter(option => option.selected)
+        .map(option => option.textContent)
+    : [htmlElement.value]
+}
+
+function getExpectedValues(expectedValue) {
+  return expectedValue instanceof Array ? expectedValue : [expectedValue]
+}
+
+function getQtyOfMatchesBetweenArrays(arrayBase, array) {
```
gnapse, line 60: I'd rather have a fully readable name like `getNumberOfMatchesBetweenArrays`. Same for the variables above that store this value.

kentcdodds:
Based on the comment above, it doesn't sound like a breaking change to me (unless someone was doing something very odd with this assertion). I think I'd be ok releasing this as a feature.

lourenci:
> This means that this will be a breaking change, technically speaking, right? There's nothing wrong about that, just want to be sure because we need to release it as such and with a new major version.

I don't think so. There is nothing in the documentation talking about that. By the way, I thought it was working in this way.

Issue #240, opened 2020-05-05
typescript declaration for ToHaveValue doesn't allow null


timrobinson33:
- `@testing-library/jest-dom` version: 5.5.0
- `node` version: 12.16.2
- `npm` (or `yarn`) version: 6.14.4

* `react-testing-library` version: (if applicable) 10.0.4

### Relevant code or config:

```js
expect(inputElement).toHaveValue(null);
```
==> typescript warning
### What you did:

I'm using <`input type="number"`> to prompt for an optional numeric value. It seems the obvious way to handle an empty value is to use null , and certainly React doesn't give any warnings when you try to do this.

### What happened:

Typescript gives a warning because the method signature for toHaveValue doesn't allow null. using the non-null assertion operator (i.e. `expect(inputElement).toHaveValue(null!);` works fine and is the best workaround IMO.

It might be possible to set an empty string to indicate an empty value, but this would require me to declare my state variable as `number | string | null` which is definitely not desirable

### Suggested solution:

From what I can tell there are already some tests calling it with null (just google toHaveValue null and they come up at the top of the list), so I presume it is valid to do this and you just need to update the typescript declarations.

gnapse:
Yes, I agree. Though the type definitions are not in this repo, but in DefinitelyTyped. Not sure what the process is there, if we also open an issue there, or just the PR (if you're up for it that's cool BTW). But I'm fine to keep this issue here to keep track of this, because ultimately is an issue with our lib, even if we do not provide the types right here (I wish we did).

timrobinson33:
OK thanks. I'm new to contributing but since it looks like you've agreed it's a fault, and it should be simple to fix, it looks like a good one to cut my teeth on. I'll try a PR on DefinitelyTyped and reference this issue.

timrobinson33:
I see the fix has been released in https://www.npmjs.com/package/@types/testing-library__jest-dom/v/5.0.4. As far as I'm concerned, you can close this issue now

eps1lon:
Fixed in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/44541/

PR #244, opened 2020-05-07
Add toHaveDescription()

BurntCaramel:
**What**:

Adds `toHaveDescription()` matcher

**Why**:

We are able to test labels and accessible names. Almost as popular are descriptions, specified by the `aria-describedby` attribute. It would be great to test these as they often complement labels.

**How**:

- Add `toHaveDescription` matcher with tests

**Checklist**:

- [x] Documentation
- [x] Tests
- [ ] Updated Type Definitions
- [ ] Ready to be merged
      

 I will need help on how to update the type definitions. But I am happy to do so! I know TypeScript well, I’m just not sure how to update the DefinitelyTyped mega repo — do you just pull down the whole thing and make a PR there?

PR #246, opened 2020-05-07
feat: native typescript support

Meemaw:
**What**:

Add native Typescript support to `jest-dom`.

**Why**:

1. To prevent issues with unsynced/wrong type declarations, e.g. https://github.com/testing-library/jest-dom/issues/240
2. To make the developer experience better (no need to update multiple repors -- types will be created for free)
3. For type safe code

**How**:

All files were converted to Typescript and global declaration added which extends `jest` matchers:

```ts
import * as extensions from './matchers'

expect.extend(extensions)

type Extensions = typeof extensions

declare global {
  namespace jest {
    interface Matchers<R, T> extends Extensions {}
  }
}
```

**Checklist**:

- [x] Documentation
- [x] Tests
- [ ] Updated Type Definitions 💣 -- not needed anymore
- [ ] Ready to be merged
      

This PR is not ready to be merged yet, there is still some polishing to be done if we want to merge it. However, I would like to get some feedback if going into that direction is even desired.


gnapse:
We used to have this (sort of, not implemented in TypeScript, but type definitions in this package), and in #123 it was decided to move them to DefinitelyTyped because they did not work out of the box in VSCode for auto-completion and things like that.

But I very much welcome this change. Thanks for all the effort. I'd rather have the library implemented in TS so you doing this is very much welcome!

Now the question is, does this satisfies the issues raised in #123?

kentcdodds:
If the TypeScript defs are a part of the package then it will work out of the box (and arguably better).

FWIW, DOM Testing Library just moved the types into the repo: https://github.com/testing-library/dom-testing-library/pull/530

gnapse:
One more thing: we may have some PRs merging soon, and with this kinds of changes, the conflicts are usually hell. For instance, I'm considering merging #244 which is an entirely new feature, but it will block this PR having to rebase and rewrite that new matcher as well.

Given all the effort you already put into this, and how much I want this to happen, would you rather have this merged first and then we can require that PR to rewrite to TS (which does not have to be the author, I am willing to step in if needed). Or do you think it's ok and we can take care of the rebasing and rewriting in this PR?

gnapse:
> If the TypeScript defs are a part of the package then it will work out of the box (and arguably better).

Ok, that's what I suspected. This has full green light then. Not sure what to merge first then, if this or #244.

Meemaw:
@gnapse I would say merge https://github.com/testing-library/jest-dom/pull/244 first and then I'll rebase on top of it.

@kentcdodds Nice! The difference is that this is a complete rewrite to Typescript (opposed to just pulling the types into the package) so there is no overhead of editing types -- they are auto generated.

gnapse:
> I would say merge #244 first and then I'll rebase on top of it.

Thanks! Let me know if you need help with the rebase or rewrite, if you do not have time soon. I'd be more than happy to help giving this the final push.

kentcdodds:
That may happen one day, but as noted in that PR I'm not certain of it. We'll see.

.travis.yml
```diff
@@ -1,16 +1,15 @@
...
 node_js:
-  - 10.14
+  - 10.18
   - 12
   - node
-install: npm install
+install: yarn install
 script:
-  - npm run validate
+  - yarn validate
```
gnapse, line 12: Why the change to yarn? Is it needed?

Meemaw: For some reason there were problems with `npm install`. It looked like it still tried to install `@types/testing-library__jest-dom` which were wrong at some places which failed tests. I don't know why it installed them (I'm not sure why they were included since they are not a dependency anymore -- it might be travis cache/and/or lack of lockfile)

Meemaw:
@kentcdodds any idea why `tsc --noEmit` would be failing from `kcd-scripts/husky` pre-commit hook? Running `tsc --noEmit` locally in terminal passes for me, which makes me think the pre-commit one doesn't take some things into account (tsconfig.json?)

kentcdodds:
Could be a bug. It's not a feature I've used much yet so I'm not sure.

.travis.yml
Meemaw: Made some changes so yarn is no longed needed.

Meemaw:
Yeah its definitely a bug, if I manually copy local `tsconfig.json` into `node_modules/kcs-scripts/dist/config` it works.

Meemaw:
@gnapse is it possible to do alpha pre-release so I can confirm this works from few of my projects. I linked the package locally and it worked, but still better be safe than sorry.

gnapse:
To be honest, not at all sure how to go about doing an alpha release. I've always relied on automated releases. @kentcdodds can you chime in on this?

kentcdodds:
Make a branch called alpha and merge it into that branch. Then make sure Travis builds that branch and it should work.

gnapse:
@Meemaw I created the `alpha` branch and set it as the base for this PR, so let me know when you're ready and we can merge it and see how it goes.

Meemaw:
I think you should actually name in `beta` to match branches travis builds on: https://github.com/testing-library/jest-dom/blob/master/.travis.yml#L18

I expect it will make a pre-release like this: https://github.com/testing-library/react-testing-library/releases/tag/v10.0.0-beta.2

gnapse:
`beta` created and this branch changed base to it

Meemaw:
Fell free to merge. I would consider squashing it into a fix/patch commit.

gnapse:
> I would consider squashing it into a fix/patch commit.

Yes, we always do that.

jgoz:
I’m just catching up on this - apologies for the delay. But won’t this change prevent types from being globally available without some kind of import statement?

If you recall discussion from #123, TypeScript will only apply global typings automatically if they are in the `@types` folder. Otherwise, you (as a user) need to have an import statement in your code or add to your `tsconfig` `include`/`files` for types to be available.

I believe this change will be a regression there since it drops the `@types` dependency. It doesn’t matter if the types are written manually or generated by TypeScript, they will not be globally available without some kind of user configuration, which is what we were trying to solve in #123.

gnapse:
I will make up some time to checkout this version of the lib locally to make sure. Did you do that @jgoz (TBH that's the only way I know, I do not know TypeScript enough to be sure of this without actually trying it out, I only know the language but I'm rusty on all its setup kind of things like this).

BTW @kentcdodds said above the opposite

>If the TypeScript defs are a part of the package then it will work out of the box (and arguably better).

which is what I'd expect, because I normally consume libraries written natively in TypeScript and they work out of the box without the need for the `@types` dependency (or maybe they have it internally and I've never checked 🤔) Anyway, I'll come back with my findings.

gnapse:
Hmmm, I'm very unhappy to report that indeed, this seems to be a regression, which I don't understand. Isn't it supposed that if the library is in TS it should all work?

Anyway, here's what I did:

1. Took a project I'm already working on and upgraded the version of jest-dom to v5.7.0-beta.1.
2. Deleted `node_modules/` (as a safety measure, to be sure I was starting clean) and ran `npm install`
3. Made sure that with this the dependency `@types/testing-library__jest-dom` is no longer present.
4. Restarted vscode and opened a test file. Then tried to type a new assertion. There is no auto-complete available for jest-dom matchers anymore :(

Isn't there a way to make this work satisfying #123 and having the code in TS @jgoz?

jgoz:
@gnapse I don't believe it's possible, which is why we went down the DefinitelyTyped route in the first place. Writing the library natively in TypeScript has no impact on how the types are consumed by users compared to JavaScript + type definitions. The big difference is internal typings vs. DefinitelyTyped typings.

Our case is a special one because the API is used in a global context (the import statement/setup is detached from actual usage). For a normal library where something is imported and used in the same file, supplying your own types or converting to TypeScript makes sense (@kentcdodds is correct about that).

But that is not how people use this library and TypeScript does not automatically include global type definitions from random packages in `node_modules`. It will only include them from `node_modules/@types`, maybe for performance reasons, maybe to avoid conflicts/confusion/etc.

So this isn't all bad news, we have a couple of options to work with here. Instead of reverting this entire PR, we could just revert the `package.json` change and re-add the `@types` dependency. And then we could use the `.d.ts` output in this repository as the source of truth and manually synchronize it with the global typings in DefinitelyTyped. Not ideal, but it would work.

It might also be possible to simply import our internal typings _from_ the DT typings so we didn't have to keep them in sync, but that might result in circular `npm` dependencies and might need special intervention from the DT maintenance team again since it is likely against the rules.

Meemaw:
@jgoz I was thinking and had an idea that might work. Wha about a postinstall script that would move typings to types folder? We get best of both worlds.

It is possible to install packages with arugument to ignore those, but I think that is very rare.

jgoz:
@Meemaw That is an interesting idea. I think as long as the script was fairly defensive (create directories as necessary, don't overwrite files) and it always operated relative to the user's `package.json` file, it could work well.

Making it relative to the user's `package.json` is important because this library might be a transitive dependency of another package and so might get installed deeper in the `node_modules` hierarchy. But `@types` packages will only get picked up automatically if they are in the top-level `node_modules` folder, nearest to the user's `package.json` file.

gnapse:
Wow, that sound cool.

> It is possible to install packages with arugument to ignore those, but I think that is very rare.

And also, nobody can complain if they then do not have the expected type awareness support in their IDE or somewhere else.

Meemaw:
I'll try that out as soon as I find some time.

benmonro:
@gnapse I'm a bit noob-sauce on TS but trying 5.7.0-beta.1 didn't work on my project.

I added `  setupFilesAfterEnv: ['./jestSetup.ts'],` to jest.config.js and  `import "@testing-library/jest-dom";` to jestSetup.ts

but my test still complains about `.toBeInTheDocument()` unless i import jest dom in the test file.  am i missing something?

gnapse:
@benmonro this is a known issue, exactly what we're discussing above (see https://github.com/testing-library/jest-dom/pull/246#issuecomment-625940278 above).

This is precisely why we released this only as a beta release for the time being. And the immediate discussion prior to your comment was about an approach that @Meemaw is looking into to solve this.

benmonro:
gotcha ok wasn't sure but thanks for clarifying!

benmonro:
@gnapse any update on this?

gnapse:
This is waiting on checking up if this idea proposed by @Meemaw could work:

> a postinstall script that would move typings to types folder

Meemaw:
Update:
I've tried that approach and it works perfectly (on first install). It seems `postinstall` script is not triggered when you add packages or something else and then the typing folder that was initially created will get cleaned up...

gnapse:
@Meemaw @jgoz how about having [jsdoc-based type checking](https://twitter.com/addyosmani/status/1216064732257492996?s=20)? I recently had a very good experience introducing it in a project that other reasons was very difficult to take directly to TypeScript.

I'm so excited about this, that I'd be willing to have that at least as a second-best outcome from this effort. I'd work on this myself since @Meemaw already laid the groundwork here.

jgoz:
@gnapse would that be an alternative to rewriting the library in TypeScript? Certainly doable if you prefer to keep things in JS, but it won’t help us export global typings. I think DefinitelyTyped is still the best way to provide typings if we are striving for zero-configuration.

PR #248, opened 2020-05-08
chore(ci): use yarn for install

Meemaw:
It seems there are still some issues with `npm install` in travis. Old types are still being pulled from somewhere 🤷‍♂️


gnapse:
Hmmm, I'd rather have that mixed, but I guess we can deal with that later. I'll merge and see how it goes.

Issue #243, opened 2020-05-06
Ability to assert mixed-checked checkboxes


gnapse:
### Describe the feature you'd like:

See testing-library/eslint-plugin-jest-dom#43

Specifically, the difficulty to asserting that a checkbox-like element is not checked or unchecked, but in a middle third state usually known as "mixed".

### Suggested implementation:

It could be a new matcher `toBeCheckedMixed`, but I'm leaning more towards keeping it under `toBeChecked`, which could receive an optional argument `'mixed'` (specifically that string, and nothing more), and will pass if the checkbox is in this third state. Without argument it will behave as today. When passed another argument, its behavior is undefined (maybe throw an error? or maybe we should support passing to it `'true'` and `'false'` and fail or pass accordingly, and still fail with anything else passed to it?)

### Describe alternatives you've considered:

Keep using `toHaveAttribute('aria-checked', 'mixed')` but in addition to cause the linting issue linked to above, it does not cover the other way in which a checkbox can be mixed-checked: `checkboxElement.indeterminate = true` ([ref](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox)).

### Teachability, Documentation, Adoption, Migration Strategy:

Add to the documentation the new feature. It does not / should not break existing functionality.

rubenmoya:
Hi! I'd love to take on this one to compensate the fail with the `toBeReadonly` PR 😂

About the matchers I like the idea of having `toBeChecked('mixed')` instead of `toBeCheckedMixed`, but I'm not so sure with `true` and `false` as arguments, `toBeChecked('true')` would be the same as `toBeChecked` so probably would end up adding more work in the `eslint-plugin-jest-dom` package to alert the user, and so would happen with `toBeChecked('false')` and `.not.toBeChecked`, what do you think?

Also, after reading about the "mixed" state I've found out that:

> The switch role does not support the value mixed for the aria-checked attribute; assigning a value of mixed to a switch instead sets the value to false.

So we should take that into account too.

gnapse:
Thanks for your input. Yeah, it does not make much sense to add the argument for the other cases. Which makes me think if it makes sense to add it at all. Maybe we should add `toBePartiallyChecked` after all. What do you think?

About the mixed state not supported on a switch, yes, we should take that into account.

rubenmoya:
[...]

PR #249, opened 2020-05-10
feat: add toBePartiallyChecked matcher

rubenmoya:
**What**:

This PR adds a `toBePartiallyChecked` matcher

**Why**:

You can read more about the why in https://github.com/testing-library/jest-dom/issues/243

**How**:

Following the implementation of `toBeChecked` but adapting it to check less elements since `radio` and `switch` cannot have a mixed state.

**Checklist**:

- [x] Documentation
- [x] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/__tests__/to-be-partially-checked.js
+++ b/src/__tests__/to-be-partially-checked.js
@@ -108,12 +108,10 @@ describe('.toBePartiallyChecked', () => {
 
     expect(() =>
       expect(queryByTestId('aria-checkbox-invalid')).toBePartiallyChecked(),
-    ).toThrowError(
-      'only inputs with type="checkbox" or elements with role="checkbox" and a valid aria-checked attribute can be used with .toBePartiallyChecked(). Use .toHaveValue() instead',
-    )
+    ).toThrowError()
   })
 
-  test('throws when the element is not an input', () => {
+  test('throws when the element is not a checkbox', () => {
     const {queryByTestId} = render(`<select data-testid="select"></select>`)
     expect(() =>
       expect(queryByTestId('select')).toBePartiallyChecked(),
--- a/src/__tests__/to-be-partially-checked.js
+++ b/src/__tests__/to-be-partially-checked.js
@@ -1,0 +1,124 @@
+import {render} from './helpers/test-utils'
+
+describe('.toBePartiallyChecked', () => {
+  test('handles input checkbox with aria-checked', () => {
+    const {queryByTestId} = render(`
+      <input type="checkbox" aria-checked="mixed" data-testid="checkbox-mixed" />
+      <input type="checkbox" checked data-testid="checkbox-checked" />
+      <input type="checkbox" data-testid="checkbox-unchecked" />
+    `)
+
+    expect(queryByTestId('checkbox-mixed')).toBePartiallyChecked()
+    expect(queryByTestId('checkbox-checked')).not.toBePartiallyChecked()
+    expect(queryByTestId('checkbox-unchecked')).not.toBePartiallyChecked()
+  })
+
+  test('handles input checkbox set as indeterminate', () => {
+    const {queryByTestId} = render(`
+      <input type="checkbox" data-testid="checkbox-mixed" />
+      <input type="checkbox" checked data-testid="checkbox-checked" />
+      <input type="checkbox" data-testid="checkbox-unchecked" />
+    `)
+
+    queryByTestId('checkbox-mixed').indeterminate = true
+
+    expect(queryByTestId('checkbox-mixed')).toBePartiallyChecked()
+    expect(queryByTestId('checkbox-checked')).not.toBePartiallyChecked()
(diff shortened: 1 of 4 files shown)


src/to-be-partially-checked.js
```diff
@@ -0,0 +1,54 @@
...
+  const isValidInput = () => {
+    return (
+      element.tagName.toLowerCase() === 'input' && element.type === 'checkbox'
+    )
+  }
+
+  const isValidAriaElement = () => {
+    return (
+      element.getAttribute('role') === 'checkbox' &&
+      element.getAttribute('aria-checked') === 'mixed'
```
rubenmoya, line 16: This check is already done in `isPartiallyChecked` probably shouldn't be here, right?

src/__tests__/to-be-partially-checked.js
```diff
@@ -0,0 +1,124 @@
+import {render} from './helpers/test-utils'
+
+describe('.toBePartiallyChecked', () => {
+  test('handles input checkbox with aria-checked', () => {
+    const {queryByTestId} = render(`
+      <input type="checkbox" aria-checked="mixed" data-testid="checkbox-mixed" />
```
gnapse, line 6: Hmmm, I don't think we should support `aria-checked` in a `<input type="checkbox" />`. It should be supported only on non-native checkbox elements with `role="checkbox"`.

The reason being that a `<input type="checkbox" />` won't convey to the user visually that it is partially checked. The only way the native checkbox can be partially checked is when its  `indeterminate` property is set to `true`.

What do you think?

src/to-be-partially-checked.js
gnapse: Yup, I don't think this should be part of this function. This function should only check if the element has a valid role for which this check makes sense. Then the actual check with the boolean value you set the `pass` property to, that's what then determines if the element is partially checked or not.

src/__tests__/to-be-partially-checked.js
rubenmoya: I thought the same when I tested it in the broswer, as you say `indeterminate` set to `true` has a visual effect on the checkbox:

image

The only thing that made me change my mind is that in the accessibility consoleboth of them had role checkbox and checked `mixed`, but yeah, I think it should be better to check for input+indeterminate and role checkbox + aria mixed, let me change it

rubenmoya:
It should be good now, let me know if I missed something

src/__tests__/to-be-partially-checked.js
gnapse: > in the accessibility console both of them had role checkbox and checked mixed

Would you care to share more about this detail?

rubenmoya: Both inputs show the same data in the _Accessibility_ tab in the Chrome Devtools:

Input with `indeterminate`:

image

Input with `aria-checked="mixed"`:

image

You can check the [Codesandbox demo](https://codesandbox.io/s/hungry-night-iwlkp?file=/src/App.js) I've been using.

gnapse: Ok, my bad then. If the `<input type="checkbox" aria-checked="mixed" />` shows as `checked: mixed` even if `indeterminate` was not set true, then I think it was ok the first time around, and I assumed without checking what you correctly went out to check.

What do you think?

If you agree, can you easily revert the changes after my earlier review? Sorry for having made you work extra for nothing.

rubenmoya: So we trust the accessibility info instead of how it's displayed right?

gnapse: Yes, I was also asking for your opinion here. But I'm conflicted. The `<input type="checkbox" aria-checked="mixed" />` looks visually as if unchecked, but semantically (from the aria perspective) it is still partially checked.

I'm leaning towards trusting the accessibility info. If someone uses an input checkbox with `aria-checked="mixed"` they would presumably take over the rendering to reflect that using icons and stuff.

So yes, in the end I'm leaning to that (after having thought about it out loud here 😄)

rubenmoya: Yeah, I thought the same at first, since it looks unchecked, but then I thought it's just a matter of how chrome styles the checkbox.

Let's support both of them then :)

gnapse: Ok, so the last commit needs to be reverted, right? It was good the first time around, right?

rubenmoya: Yes, I'll revert it and change only the check inside `isValidAriaElement`

rubenmoya: Should be good now :)

rubenmoya:
Not sure why Travis is failing, any idea?

gnapse:
> Not sure why Travis is failing, any idea?

Not sure either. I re-launched the build and it worked. Will review soon.

gnapse:
@all-contributors add @rubenmoya for code, test, docs

PR #250, opened 2020-05-13
feat: do not explicitly depend on jest assertion utils

SimenB:
**What**:

Replaced the direct usage of `jest-matcher-utils` and `jest-diff` with using the same functions from `this.utils` injected into each matcher.

**Why**:

I noticed I had old `jest-diff` in my `node_modules` after upgrading to v26 and decided to fix at least one of the entries 🙂 

This change allows `jest-dom` to not depend on explicit versions of these dependencies, instead getting the utils from `expect` during runtime.

**How**:

Search and replace, followed by making the linter happy :grinning:

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged


README.md
```diff
@@ -46,7 +46,6 @@ clear to read and to maintain.
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 
-
```
SimenB, line 49: I just ran `yarn format` :shrug:

eps1lon:
> This change allows jest-dom to not depend on explicit versions of these dependencies, instead getting the utils from expect during runtime.

So we'd need a peer dependency on `jest`, no?

SimenB:
Not, it's `expect` that injects these, not Jest. You already call `expect.extend` without a peer dep on it. And you already depend on e.g. `this.isNot` so in that regard this PR changes nothing

SimenB:
@gnapse any holdup here?

gnapse:
Nope, my bad. Will merge soon.

gnapse:
@SimenB there have been a couple of new matchers since you proposed this change. I can take care of it later if you don't beat me to it.

PR #251, opened 2020-05-13
Remove TypeScript warning from README

MoSattler:
This does not see to be the case anymore, TS works fine since #11

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,22 +120,6 @@ import '@testing-library/jest-dom'
 > Note: If you're using TypeScript, make sure your setup file is a `.ts` and not
 > a `.js` to include the necessary types.
 
-Alternatively, you can selectively import only the matchers you intend to use,
-and extend jest's `expect` yourself:
-
-```javascript
-import {
-  toBeInTheDocument,
-  toHaveClass,
-} from '@testing-library/jest-dom/matchers'
-
-expect.extend({toBeInTheDocument, toHaveClass})
-```
-
-> Note: when using TypeScript, this way of importing matchers won't provide the
-> necessary type definitions. More on this
-> [here](https://github.com/testing-library/jest-dom/pull/11#issuecomment-387817459).
-
 ## Custom matchers
 
 `@testing-library/jest-dom` can work with any library or framework that returns


gnapse:
I'm not sure that this is the case.

When you import some matchers manually and not others, you need to extend `expect` yourself. And therefore, you need to provide the extend types for custom matchers yourself too, providing the types for only those that you manually imported and extended `expect` with.

Or am I missing something here?

MoSattler:
Can you give a concrete example where typing is not properly working ATM?

gnapse:
Ok, yes, you are right. Now I realized we have a different problem. If someone uses the custom matchers in this way, then they now have type definitions for all matchers, even those they did not import. They get auto-complete in jest-dom even.

I wonder if we should remove this section altogether (the reference to importing custom matchers and extending `expect` manually). It does not seem much useful. I don't even know what the motivation would be to do so. It's not like these custom matchers are contributing to your bundle size or anything.

Would like to hear others' thoughts about this.

MoSattler:
@gnapse any news on this?

gnapse:
Oh yes, sorry. Would you mind removing the entire section altogether? From where we start saying "Alternatively you can import only the matchers you need" (something along those lines). It's line 121 I think.

If not don't worry, we can also merge this as is. Let me know.

Grmiade:
Typings is not properly working at the moment on my side with this example:

```typescript
import {
  toBeInTheDocument,
  toHaveClass
} from "@testing-library/jest-dom/matchers";

expect.extend({ toBeInTheDocument, toHaveClass });
```

```
Could not find a declaration file for module '@testing-library/jest-dom/matchers'. '/Users/jeremy/Development/backoffice-web-client/node_modules/@testing-library/jest-dom/matchers.js' implicitly has an 'any' type.
  Try `npm install @types/testing-library__jest-dom` if it exists or add a new declaration (.d.ts) file containing `declare module '@testing-library/jest-dom/matchers';`
```

https://codesandbox.io/s/cocky-microservice-qwjx5?file=/src/setupTests.ts

nickmccurdy:
Are you sure you have at least version `1.1.0` installed of `@testing-library/jest-dom`? This works fine for me with the latest version of `jest-dom` and `typescript`.

gnapse:
The thing is we really do not support that and TypeScript at the same time. Not sure we can. My take on this was that we should remove this section altogether from the README. That's what I last requested from @MoSattler. Sadly I can't add changes to this PR, but I can start a new one superseding this one.

MoSattler:
@gnapse sorry for the delay. Did as you requested!

gnapse (approved):

gnapse:
@all-contributors add @MoSattler for docs

Issue #216, opened 2020-03-17
ToBeEmpty assertion in jest-dom is also used in jest-extended


DanielaValero:
### Describe the feature you'd like:
Jest-Extended has an assertion named the same as jest-dom's one:

-> https://github.com/jest-community/jest-extended#tobeempty
-> https://github.com/testing-library/jest-dom#tobeempty

When both libs are used in the same project, the version of `jest-extended` overwrites the one from jest-dom.

### Describe alternatives you've considered:
Maybe an option is to add an alias  for `toBeEmpty` named `toBeEmptyNode` ?

### Teachability, Documentation, Adoption, Migration Strategy:
I am not sure at this moment how the things work, could it be an option to have 2 names for the same assertion?

If yes, then the documentation could say: 
"if you are using jest-extended, then use `toBeEmptyNode`" or something like this

**edit**
 I have raised a feature request in Jest Itself asking for a feat that warns the user about the name collision and points out to a link where is documented a way to rename one of the colliding names in the project where the situation happens.

https://github.com/facebook/jest/issues/9678

It could be that the warning and the documentation of how to solve it in projects, might be enough to address this and similar situations like this one.

gnapse:
I noticed this a while ago and even brought it up as an issue with jest: facebook/jest#6243

I could see us providing an alias as you suggest.

darekkay:
I've also encountered this issue today after introducing `jest-dom` into a large project. 

> When both libs are used in the same project, the version of jest-extended overwrites the one from jest-dom.

This depends on the import order. In my case, it's `jest-dom` that overrides `jest-extended`. So the proposed solution to add an alias still requires the correct import order.

Another way to solve it is to actually support other "empty" checks, similar to what [lodash](https://lodash.com/docs/4.17.15#isEmpty) considers "empty". As `jest-dom` concentrates on the _DOM_ aspect, the proposed alias solution seems to be a better one, though.

As a workaround, one can write a custom, third `toBeEmpty` extension that uses one of the library implementation based on the object type.

gnapse:
I honestly don't know how to solve this properly. IMO it lies on the jest side of things. Can you all at least comment in the linked issue in jest (facebook/jest#6243) so that this gets some visibility: I doubt it could have a quick fix, but this is something they should address. and there are mentions there of proposed solutions and maybe even a workaround?

gnapse:
Though I'd be open to add an alias `toBeEmptyDOMElement` or something along those lines.

[...]

PR #254, opened 2020-05-24
feat: Deprecate toBeEmpty in favour of toBeEmptyDOMElement (#216)

DanielaValero:
**What**:
- Add deprecation note to `toBeEmpty`
- Create `toBeEmptyDomElement` to replace `toBeEmpty`
- Document changes

**Why**:
This is an alternative solution that allows users of `jest-dom` and `jest-extended` to check for emptiness of strings, etc. And DOM elements, using different matchers.

**How**:
- The code of `toBeEmptyDomElement` is a copy of `toBeEmpty`, the only changes in there is the name.

**Checklist**:
- [x] Documentation
- [x] Tests
- [N/A] Updated Type Definitions
- [?] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ clear to read and to maintain.
   - [`toBeDisabled`](#tobedisabled)
   - [`toBeEnabled`](#tobeenabled)
   - [`toBeEmpty`](#tobeempty)
+  - [`toBeEmptyDOMElement`](#tobeemptydomelement)
   - [`toBeInTheDocument`](#tobeinthedocument)
   - [`toBeInvalid`](#tobeinvalid)
   - [`toBeRequired`](#toberequired)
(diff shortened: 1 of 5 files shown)


DanielaValero:
Hi @gnapse so should I leave it with this name? I could do another PR to remove `toBeInTeDOM`

What is the usual process to merge a PR? one approval and I merge myself? two approvals and the second approver merges?

gnapse:
I'd rename this to `toBeEmptyDOMElement` regardless of the future removal of `toBeInTheDOM`. As @KevinHerklotz pointed out above, it is a better reflection of the words in the phrase.

@DanielaValero the removal of `toBeInTheDOM` is an independent issue. If you still want to take a crack at it, please do so, but do not do it to remove the discrepancy only. And thanks a lot!

DanielaValero:
Hi, about the renaming, sure, it took me a while to notice the part needs renaming is to make DOM all uppercase. 

If I might suggest yo an improving area when giving feedback, in the case of renaming or typos, authors of the code not necessarily notice when they did the typo or renaming bit, and it helps a lot to point out specifically what is the part. i.e: Dom -> DOM 🙂

As of the `toBeInTheDom` I suggested I could do it in a separate PR definitely not with the goal of removing the discrepancy, I meant it more to offer help because you mentioned can be removed.

Thanks for the clarification on the feedback!

gnapse:
@DanielaValero sorry. Yes, I assumed it was evident when it is not. Will keep that in mind. I realize I never mentioned the fact that this word is an acronym even, which would've given the clue about the all-caps vs. capitalized difference.

> I meant it more to offer help because you mentioned can be removed.

Sure. Your help is greatly appreciated.

I'll review it all shortly. Thanks!

gnapse:
@all-contributors add @DanielaValero for code, test, docs

charliematters:
Is there an ETA on this arriving in the @types repo?

darekkay:
@gnapse @charliematters The [types PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45149) is merged and a new [package version](https://www.npmjs.com/package/@types/testing-library__jest-dom) is released, so we only have to update the version [here](https://github.com/testing-library/jest-dom/blob/master/package.json#L35), right? If yes, I can provide a quick PR.

gnapse:
Yes, please. Do so. Thanks!

Issue #256, opened 2020-06-03
toHaveErrorMessage


jacobparis:
### Describe the feature you'd like:

I'd like an assertion for the `aria-errormessage` attribute. It accepts a list of id strings formatted identically to the format `aria-describedby` accepts.

https://www.w3.org/TR/wai-aria-1.1/#aria-errormessage

Browser support was poor a year ago but I can't seem to find any up-to-date information on its support level. There is *some* chance that this is unsupported enough to be not worth implementing, but on a long enough timeline I think it'll fit well here.

Usage
```html
<input
  type="text"
  id="mortgage-amount"
  name="MORTGAGE_AMOUNT"
  aria-label="Mortgage Amount"
  aria-describedby="mortgage-amount-description"
  aria-invalid="true"
  aria-errormessage="mortgage-amount-error"
/>

<p id="mortgage-amount-error"> Must be a positive number </p>
````
```js
expect().toHaveErrorMessage('Must be a positive number')
```

### Suggested implementation:

The implementation would be nearly a clone of `toHaveDescription` except the elements it reads come from the `aria-errormessage` attribute instead of `aria-describedby`

It could also assert that `aria-invalid` is true

I'm willing to make a PR for inclusion but I'm new to open source so I wanted to get feedback up front

### `toHaveErrorMessage`

```typescript
toHaveErrorMessage(text: string | RegExp)
```

This allows you to check whether the given element has an error message or not.

An element gets its error message via the
[`aria-errormessage` attribute](https://www.w3.org/TR/wai-aria-1.1/#aria-errormessage).
Set this to the `id` of one or more other elements. These elements may be nested
inside, be outside, or a sibling of the passed in element.

Whitespace is normalized. Using multiple ids will
[join the referenced elements’ text content separated by a space](https://www.w3.org/TR/accname-1.1/#mapping_additional_nd_description).

When a `string` argument is passed through, it will perform a whole
case-sensitive match to the error message text.

To perform a case-insensitive match, you can use a `RegExp` with the `/i`
modifier.

To perform a partial match, you can pass a `RegExp` or use
`expect.stringContaining("partial string")`.

#### Examples

```html
<!-- with aria-errormessage and aria-invalid -->
<input type="text" aria-label="Username" aria-errormessage="username-error" aria-invalid="true"/>

<div id="username-error">
  Username must not contain spaces
</div>

<!-- with just aria-errormessage -->
<input type="text" aria-label="Email" aria-errormessage="email-error" />

[...]

PR #258, opened 2020-06-03
Adds toHaveErrorMessage assertion

JacobParis:
**What**:

I implemented a basic toHaveErrorMessage assertion. It currently does not check for visibility (suggested by spec)

Original issue: https://github.com/testing-library/jest-dom/issues/256

I'd like some recommendations for how to go about adding those two missing features. The logic to tell whether `aria-invalid` is set is already present in the `toBeInvalid` assertion and `isVisible` is in a similar boat.

My first reaction would be to reuse that functionality here, but as of right now there is no precedent in the codebase for one assertion to call the code of another.

I could re-implement it in this handler, but then there's two branches to maintain in parallel for the same functionality.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Ready to be merged


gnapse:
I don't think what we need to check here is exactly all that `isValid`/`isInvalid` already check. The spec mentions specifically that `aria-invalid` is the key, and those other matchers cover other type of validity checking as well. The check needed here is simpler, merely check that the same element where `aria-errormessage` attribute is found must also have the `aria-invalid="true"` attribute present **and** set to `true`. That's as simple as what this function already does:

https://github.com/testing-library/jest-dom/blob/159a7a263acc1c7af3f29dccfe70cfc85ea05bba/src/to-be-invalid.js#L6-L11

I'm hesitant to extract such a simple functionality to a helper in utils, so feel free to copy the equivalent condition in that code to the file implementing this new assertion.

As for the other part, I also do not think that `toBeVisible` covers the visibility check we need here. It tests for actual "seeing it with your eyes" visibility, whereas with this one I suspect that screen-reader visibility is enough (e.g. making sure the element or one of its ancestors in the DOM tree does not have `display: none` and maybe not having also `visibility: hidden`).

However as we agreed on the discussion in the issue where you proposed this, I'd be ok if you omit that part, and we can add it when it becomes a problem. I'm not sure how to fully interpret the spec on this regard correctly. And also given that visibility is a tricky issue as it may be affected by the presence of stylesheets loaded in the document, something that usually in the context of jest tests of isolated components does not happen. We already have our fair share of issue reports of people complaining that matchers such as `toBeVisible` pass when they shouldn't, and it is for instance in cases where their components apply a `class="hidden"` to elements, but if that css class definition is not loaded and attached to the document, it does not indicate the element is hidden from jest-dom's POV.

JacobParis:
Thank you for the detailed response. I think that clarifies all my concerns

JacobParis:
I'm pretty happy with how this is looking right now. I could use some feedback on the test failure messages, and once we're all good with that I can move on to the documentation and types

Screen Shot 2020-06-15 at 10 19 14 PM
Screen Shot 2020-06-15 at 10 21 09 PM

gnapse:
@JacobParis sorry for having dropped the ball on this one for so many days.

Those error messages look good to me. I updated this branch with the latest from master. Let's see how the CI behaves this time, and if it all passes I'll do a final review and approve and merge if everything's ok.

php-coder:
@JacobParis Hi, could you, please, rebase this PR again to give @gnapse a chance to review and merge it?

SevenOutman:
Any updates on this PR? I believe it's the correct way to assert form validation instead of toHaveDescription.

gnapse:
The CI checks have failures. There's an issue with the test coverage. I believe you have to add tests to cover for some lines that are not being tested. This often is related to there being no tests for the negative assertions. For instance, this often happens when you test that both `toHaveErrorMessage` and `not.toHaveErrorMessage` are working as expected, but you also have to add tests for when `toHaveErrorMessage` throws an error because indeed there was no error message, **AND** you also need to test that `not.toHaveErrorMessage` throws when there is indeed an error message present.

It may help that you run `npm run test:update` which will generate locally a coverage report in the `coverage/` folder. Open `coverage/index.html` and you'll be able to see what the uncovered lines of code are.

SevenOutman:
Friendly ping. @JacobParis

SevenOutman:
Friendly ping. @JacobParis

gnapse:
Superseded by #370.

Issue #252, opened 2020-05-18
Allow aria-invalid on any element?


BurntCaramel:
Currently the `toBeValid` matcher whitelists what elements are allowed:
https://github.com/testing-library/jest-dom/blob/5c9e8e5a1a18e58c0bf8147baaf568259c3c2a13/src/to-be-invalid.js#L4

However, AFAIK the ARIA specification allows it on any element:

https://www.w3.org/TR/html-aria/#IDL-ValidityState
> The aria-invalid attribute may be used on any HTML element that allows global aria-* attributes except for a submittable element that does not satisfy its validation constraints.

I propose removing the element tag check and allowing it on any element.

My specific issue is with a `ol` with role `radiogroup` that I am setting `aria-invalid` on.

gnapse:
Yes, I agree. The script element type check should be only when considering the element's `checkValidity()` return value. But when checking for the presence of `aria-invalid` any element is ok.

BurntCaramel:
How would you check if it supports `checkValidity()` — `'checkValidity' in element`?

I am happy to work on this!

gnapse:
> How would you check if it supports checkValidity()

The part that checks what `.checkValidity()` returns should only be used if the element's type is one of those listed in the `FORM_TAGS` array. If it's not one of those, the only check that should be tested is the one about the presence of `aria-invalid`.

> I am happy to work on this!

Go for it! And thanks! Thanks a lot.

PR #259, opened 2020-06-11
Allow aria-invalid on any element

missilev:
Closes https://github.com/testing-library/jest-dom/issues/252

**What**:
This PR allows checking if any element is valid, not only on form elements.

**Why**:
Regarding #252 we need to provide a possibility to check the validity of the element even it's not a form element.

**How**:
I adapted `isElementInvalid` function to conditionally call `checkValidity` if element supports it.

**Checklist**:

- [x] Documentation
- [x] Tests
- [ ] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,10 +274,9 @@ expect(
 toBeInvalid()
 ```
 
-This allows you to check if a form element, or the entire `form`, is currently
-invalid.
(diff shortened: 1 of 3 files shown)


gnapse:
@all-contributors add @missilev for code, test

PR #261, opened 2020-06-12
Fix deep element not allowed to be disabled being returned as disabled

lourenci:
Elements not allowed to be disabled are returning as disabled when inside a disabled fieldset.

```
<fieldset disabled>
  // <a /> is enabled even when inside a disabled fieldset
  <a />
</fieldset>
```

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged


gnapse:
@all-contributors add @lourenci for code, test, bug

PR #267, opened 2020-06-25
Get supported roles for toBeChecked via aria-query

### LINKED ISSUE #266: Support menuitemcheckbox/menuitemradio roles in toBeChecked

According to the spec [`menuitemcheckbox`](https://www.w3.org/TR/2017/REC-wai-aria-1.1-20171214/#menuitemcheckbox) and [`menuitemradio`](https://www.w3.org/TR/2017/REC-wai-aria-1.1-20171214/#menuitemradio) both require `aria-checked` attributes.

I think it would be nice to be able to write:

```
expect(getByRole('menuitemradio', {name: 'Some option'})).toBeChecked();
```

### Suggested implementation:

Add `menuitemcheckbox` and `menuitemradio` to the [allowed roles in the matcher](https://github.com/testing-library/jest-dom/blob/0428c5c16ad80783fe0a033532c6dbd5436ff149/src/to-be-checked.js#L16).

So far, I was only able to use low level matchers like `toHaveAttribute('aria-checked', 'true')`.

The roles would need to be added to list of allowed roles mentioned in [the readme](https://github.com/testing-library/jest-dom#tobechecked) and [the error messages](https://github.com/testing-library/jest-dom/blob/0428c5c16ad80783fe0a033532c6dbd5436ff149/src/to-be-checked.js#L25).

Happy to prepare a PR.

tf:
Instead of hard coding a list of roles, use information provided by the
`aria-query` package to determine whether an element with a given role
supports the `aria-checked` attribute.

Dynamically generate an error message listing all supported roles. Make
error message expectations more fuzzy to prevent test failures if the list
of supported roles should ever change.

Fixes #266

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
   "dependencies": {
     "@babel/runtime": "^7.9.2",
     "@types/testing-library__jest-dom": "^5.9.1",
+    "aria-query": "^4.2.2",
     "chalk": "^3.0.0",
     "css": "^2.2.4",
     "css.escape": "^1.5.1",
@@ -55,7 +56,9 @@
     },
     "overrides": [
       {
-        "files": ["src/__tests__/*.js"],
+        "files": [
+          "src/__tests__/*.js"
+        ],
         "rules": {
           "max-lines-per-function": "off"
         }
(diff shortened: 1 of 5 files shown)


gnapse:
@all-contributors add @tf for code, test, idea

PR #284, opened 2020-07-27
Suggest using toBeEmptyDOMElement instead of toBeEmpty

brrianalexis:
**What**:

Suggest using toBeEmptyDOMElement instead of toBeEmpty.

**Why**:

Since PR #254 we are deprecating toBeEmpty in favour of toBeEmptyDOMElement, so we should suggest using it instead.

**How**:

When checking with an empty string using toHaveTextContent, the following error message is displayed:
```
Checking with empty string will always match, use .toBeEmpty() instead
```

In this PR I changed it to:
```
Checking with empty string will always match, use .toBeEmptyDOMElement() instead
```

**Checklist**:

- [ N/A] Documentation
- [x] Tests
- [ N/A] Updated Type Definitions
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/to-have-text-content.js
+++ b/src/__tests__/to-have-text-content.js
@@ -89,6 +89,6 @@ describe('.toHaveTextContent', () => {
 
     expect(() =>
       expect(container.querySelector('span')).toHaveTextContent(''),
-    ).toThrowError(/toBeEmpty()/)
+    ).toThrowError(/toBeEmptyDOMElement\(\)/)
   })
 })
--- a/src/__tests__/to-have-text-content.js
+++ b/src/__tests__/to-have-text-content.js
@@ -83,7 +83,7 @@ describe('.toHaveTextContent', () => {
     )
   })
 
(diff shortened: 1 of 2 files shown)


src/__tests__/to-have-text-content.js
```diff
@@ -83,7 +83,7 @@ describe('.toHaveTextContent', () => {
     )
   })
 
-  test('when matching with empty string and element with content suggest using toBeEmpty instead', () => {
+  test('when matching with empty string and element with content, suggest using toBeEmptyDOMElement instead', () => {
```
gnapse, line 86: I wonder why the `toThrowError` line below did not have to be changed as well. Shouldn't the expected error message change to `/toBeEmptyDOMElement()/`?

I think it's because there's a mistake in that regular expression. Those parenthesis are not being matched but they are interpreted as an empty group. Can you please change that regexp to be `/toBeEmptyDOMElement\(\)/`?

brrianalexis: Nice catch. It slipped past me.

I just changed it. Thanks!

PR #286, opened 2020-08-06
fix: add types to package.json

### LINKED ISSUE #245: Typing problems
Hi, i am having some typing issues

- `@testing-library/jest-dom` version:
- `node` version: 12.14.1
- `npm` (or `yarn`) version: 6.13.4
- `WebStorm 2020.1.1`
- `Next 9.3.5`

### Relevant code or config:
I am trying to use the **toHaveTextContent** function. If I run the test it works correctly, but at the typing level it keeps telling me that there are problems

tag

### What confuses me is that the types exist, but he is not taking them.

types

### What you did:

Here I show you my settings:

babel config

jest config

jest setup

package

```  
{
  "compilerOptions": {
    "target": "es5",
    "lib": [
      "dom",
      "dom.iterable",
      "esnext"
    ],
    "allowJs": true,
    "skipLibCheck": true,
    "strict": true,
    "forceConsistentCasingInFileNames": true,
    "noEmit": true,
    "esModuleInterop": true,
    "module": "esnext",
    "moduleResolution": "node",
    "resolveJsonModule": true,
    "isolatedModules": true,
    "jsx": "preserve"
  },
  "exclude": [
    "node_modules"
  ],
  "include": [
    "next-env.d.ts",
    "**/*.ts",
    "**/*.tsx",
    "jest.setup.ts"
  ],
  "typeRoots": [
    "node_modules/@types"
  ]
}
``` 

### Problem description:

any suggestion?

timdeschryver:
**What**:

Closes #245

**Why**:

**How**:

Adds a reference to the type definitions inside the `package.json`

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged
      

After I made this change locally, it could find the matchers:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "0.0.0-semantically-released",
   "description": "Custom jest matchers to test the state of the DOM",
   "main": "dist/index.js",
+  "types": "extend-expect.d.ts",
   "engines": {
     "node": ">=8",
     "npm": ">=6",


timdeschryver:
🤔 After a second attempt it just seemed to work (even without the suggested change).
My first install could perhaps had some issues?

That said, I still think adding the types to the package is can be added.
Feel free to close this tho, as it seems to just work.

gnapse:
There are unresolved paths of the conversation in that original issue report. I am not convinced the change here is a proper fix, because the intent with that `extend-expect.*` files is to remove them eventually in an upcoming major release, now that we can import directly from the package (e.g. `import '@testing-library/jest-dom';`)

I would appreciate being able to check out a minimal repo where that original issue can be reproduced.

timdeschryver:
That seems fair @gnapse , thanks!
I'll close this PR because it seems to be an hiccup on my end, which is now resolved.

PR #288, opened 2020-08-11
toHaveStyle() is still broken in 5.11.3

andy128k:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/__tests__/to-have-style.js
+++ b/src/__tests__/to-have-style.js
@@ -245,4 +245,14 @@ describe('.toHaveStyle', () => {
       })
     })
   })
+
+  test('position -9999', () => {
+    const {container} = render(`
+      <input type='submit' style="position: absolute; left: -9999px;" />
+    `)
+    expect(container.querySelector('input')).toHaveStyle({
+      position: 'absolute',
+      left: -9999,
+    })
+  })
 })


gnapse:
I am not sure why the outcome is different than with `font-size`. A workaround, and what I'd recommend, is that you match against the actual string of the `left` property value:

```js
  test('position -9999', () => {
    const {container} = render(`
      <input type='submit' style="position: absolute; left: -9999px;" />
    `)
    expect(container.querySelector('input')).toHaveStyle({
      position: 'absolute',
      left: '-9999px',
    })
  })
```

andy128k:
@gnapse 

Now I see that my test was not working properly from the beginning. `{ position: 'absolute', left: -9999 }` was internally converted to `{ position: 'absolute', left: '' }`. So, no real comparison happened.

In other words, this somehow passed before 5.11.1 :(

```
  test('position -9999', () => {
    const {container} = render(`
      <input type='submit' style="position: absolute; left: -9999px;" />
    `)
    expect(container.querySelector('input')).toHaveStyle({
      position: 'absolute',
      left: 123,
    })
  })
```

Now it fails as it should.

PR #298, opened 2020-10-13
fix: treat shadow DOM elements as in the document

geoffrich:
**What**:

`expect(element).toBeInTheDocument()` would return false when `element` was in a shadow root. It now returns true.

**Why**:
I want `toBeInTheDocument` to treat elements in a shadow root as in the document. I am experimenting with using Testing Library on apps that use custom elements.

**How**:
`toBeInTheDocument` originally checked that `element.ownerDocument.contains(element)` was true. However, this returns false if the element being checked is within a shadow root.

I changed the function to use [getRootNode](https://developer.mozilla.org/en-US/docs/Web/API/Node/getRootNode) instead.  `getRootNode` will return the element's root (usually either the document or the element's shadow root). Passing `{composed: true}` will return the document instead of any shadow roots, so we can compare this to `element.ownerDocument` to see if the element is in the document.

All existing tests pass with these changes, but let me know if you can think of any edge cases I missed.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged
      

I am not sure what the browser support requirements are for this library, but `getRootNode` does not work in IE11 and non-Chromium Edge. If that's a problem I can work on an alternate approach.

### DIFF
--- a/src/__tests__/to-be-in-the-document.js
+++ b/src/__tests__/to-be-in-the-document.js
@@ -2,12 +2,29 @@ import {HtmlElementTypeError} from '../utils'
 import document from './helpers/document'
 
 test('.toBeInTheDocument', () => {
+  const window = document.defaultView
+
+  window.customElements.define(
+    'custom-element',
+    class extends window.HTMLElement {
+      constructor() {
+        super()
+        this.attachShadow({mode: 'open'}).innerHTML =
+          '<div data-testid="custom-element-child"></div>'
+      }
+    },
+  )
+
   document.body.innerHTML = `
     <span data-testid="html-element"><span>Html Element</span></span>
-    <svg data-testid="svg-element"></svg>`
+    <svg data-testid="svg-element"></svg>
+    <custom-element data-testid="custom-element"></custom-element>`
 
   const htmlElement = document.querySelector('[data-testid="html-element"]')
   const svgElement = document.querySelector('[data-testid="svg-element"]')
+  const customElementChild = document
+    .querySelector('[data-testid="custom-element"]')
+    .shadowRoot.querySelector('[data-testid="custom-element-child"]')
   const detachedElement = document.createElement('div')
   const fakeElement = {thisIsNot: 'an html element'}
   const undefinedElement = undefined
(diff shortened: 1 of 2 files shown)


gnapse (approved):
@geoffrich nice idea!

It looks good, and looks safe to me.

nickmccurdy:
Ready to merge?

gnapse:
yup, I forgot about it.

One thing: I'm not 100% this is a patch release (a fix). Isn't this potentially breaking tests to others that may have had stuff under shadow DOM that were not found? (apologies if this makes little sense, I'm not 100% familiar with shadow dom)

nickmccurdy:
That depends on whether we consider not discovering shadow DOM elements as a bug or missing feature from the previous version.

geoffrich:
It's possible that someone has a test depending on this behavior, but I think it's unlikely. Testing Library queries do not currently return any elements in the shadow DOM -- they use `document.querySelectorAll` under the hood, which only returns elements in the light DOM. So you would have to query the shadow root directly (not using any testing library queries) to get the element you're looking for, and then expect it not to be in the document. 

Here's an example with the custom element I defined in the tests:

```js
const customElement = document.querySelector('custom-element');
const shadowDiv = customElement.shadowRoot.querySelector('div');
expect(shadowDiv).not.toBeInTheDocument(); // passed before my commit, and will now fail
```

So it's potentially a breaking change for that kind of test, but that test seems contrived -- what's the point of expecting the shadowDiv to not be in the document? It was a brittle test, because it would have also passed if querySelector returned nothing. If they were doing something like this, it's more likely that they _would_ expect shadowDiv to be in the document (as I did) and be confused when it didn't work.

That was my rationale for treating this as a bug and not a new feature. However, I don't care too much one way or another, so I can reword the commit if you like.

nickmccurdy:
It is a fairly specific example and I agree it's contrived, so it sounds like a bug to me, and therefore not a breaking change. I'm interested in what other maintainers think about it though.

gnapse:
@all-contributors please add @geoffrich for code, test, idea, bug

PR #304, opened 2020-11-02
Improved error messaging for toHaveFocus assertion

syneva-runyan:
**What**:

Added more details to the .toHaveFocus assertion error message

**Why**:
The DOM output when an element is not properly focus does not show any meaningful hints to why an assertion may have failed, which can make errors more difficult to debug 

Screen Shot 2020-11-02 at 10 54 14 AM

**How**:
Added additional verbiage to the error message
Screen Shot 2020-11-02 at 10 53 47 AM

**Checklist**:

- [x] Documentation
- [x] Tests
- x] Updated Type Definitions
- [x] Ready to be merged

Open to feedback on the messaging content or verbiage!

### DIFF
--- a/src/to-have-focus.js
+++ b/src/to-have-focus.js
@@ -15,7 +15,7 @@ export function toHaveFocus(element) {
         '',
         'Expected element with focus:',
         `  ${this.utils.printExpected(element)}`,
-        'Received element without focus:',
+        'Received element with focus:',
         `  ${this.utils.printReceived(element.ownerDocument.activeElement)}`,
       ].join('\n')
     },
--- a/src/to-have-focus.js
+++ b/src/to-have-focus.js
@@ -13,9 +13,9 @@ export function toHaveFocus(element) {
           '',
         ),
         '',
-        'Expected',
+        'Expected element with focus:',
         `  ${this.utils.printExpected(element)}`,
-        'Received:',
+        'Received element without focus:',
         `  ${this.utils.printReceived(element.ownerDocument.activeElement)}`,
       ].join('\n')
     },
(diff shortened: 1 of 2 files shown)


src/to-have-focus.js
```diff
@@ -13,9 +13,9 @@ export function toHaveFocus(element) {
           '',
         ),
         '',
-        'Expected',
+        'Expected element with focus:',
         `  ${this.utils.printExpected(element)}`,
-        'Received:',
+        'Received element without focus:',
```
gnapse, line 18: The element printed on this part of the error message is not an element without focus. It is the element that currently has the focus. So this proposed new heading is incorrect.

gnapse (changes requested):
Thanks for your contribution. I can see how the current output is not super helpful.

However, what you propose is not accurate. Probably the same confusing nature of this message lead you to misinterpret it. But the fact is that what we print in the part of the error labeled with "Received" is the element that had the focus. It is meant to let you know where the focus lied instead of where the test expected.

In the case of your screenshot the focus lies in the body element, which is what happens usually when the element that had the focus was removed or hidden from being rendered.

src/to-have-focus.js
syneva-runyan: Ahh gotcha.  Updated the error message to better reflect

syneva-runyan:
@gnapse  thanks for the feedback!  Updated the verbiage to read:

Expected element with focus:"...
Received element with focus:"...

gnapse:
@syneva-runyan sorry, the other day I noticed that the build is failing, and forgot to notify you. I took a look at the build on travis and it's not clear why it failed. A CI process stopped after 10 minutes without output or something like that. I also don't find a way to request travis to restart the build. I wonder if you can push a change (maybe an extra empty commit) to see if a new build is launched.

syneva-runyan:
@gnapse  no worries, just pushed some commits and looks like it's running now

gnapse (approved):

gnapse:
@all-contributors add @syneva-runyan for code

PR #307, opened 2020-11-11
fix: check for any Node in toHaveTextContent

### LINKED ISSUE #306: toHaveTextContent should support any Node
- `@testing-library/jest-dom` version: 5.11.5
- `node` version: 12.18.0
- `yarn` version: 1.22.5
- `react-testing-library` version: 11.1.0

### Relevant code or config:

```js
    const resourceLabel = getByText('Resource:');
    const valueElement = ensureExists(resourceLabel.nextSibling);
    expect(valueElement).toHaveTextContent('libxul.so');
```

### What you did:

I'm in the process of installing and using jest-dom in our project. I installed the eslint plugin which converted this code:
```js
    const resourceLabel = getByText('Resource:');
    const valueElement = ensureExists(resourceLabel.nextSibling);
    expect(valueElement.textContent).toBe('libxul.so');
```

### What happened:
Here is the error I get:
```
    Received has type:  object
    Received has value: libxul.so

       95 |     const resourceLabel = getByText('Resource:');
       96 |     const valueElement = ensureExists(resourceLabel.nextSibling);
    >  97 |     expect(valueElement).toHaveTextContent('libxul.so');

```

### Reproduction:
https://codesandbox.io/s/react-testing-library-demo-forked-0puq7

### Problem description:

The property `textContent` is present in all `Node`s, not only `HTMLElement`. Therefore it would be nice if this expectation worked on all `Node`s too, especially text nodes.

See the MDN page about `textContent`:
https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent

### Suggested solution:
Is removing this line enough:
https://github.com/testing-library/jest-dom/blob/c11f9c53bbf94c1aa48b401f4f523500e904e128/src/to-have-text-content.js#L8

Or should we replace it with a check for `instanceof Node` ?

nickmccurdy:
**What**:

Fixes #306

**How**:

`!(htmlElement instanceof Node)`

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/to-have-text-content.js
+++ b/src/to-have-text-content.js
@@ -1,11 +1,15 @@
-import {checkHtmlElement, getMessage, matches, normalize} from './utils'
+import {getMessage, HtmlElementTypeError, matches, normalize} from './utils'
 
 export function toHaveTextContent(
   htmlElement,
   checkWith,
   options = {normalizeWhitespace: true},
 ) {
-  checkHtmlElement(htmlElement, toHaveTextContent, this)
+  const window = htmlElement.ownerDocument.defaultView
+
+  if (!(htmlElement instanceof window.Node)) {
+    throw new HtmlElementTypeError(toHaveTextContent, this)
+  }
 
   const textContent = options.normalizeWhitespace
     ? normalize(htmlElement.textContent)
--- a/src/__tests__/to-have-text-content.js
+++ b/src/__tests__/to-have-text-content.js
@@ -10,6 +10,14 @@ describe('.toHaveTextContent', () => {
     expect(queryByTestId('count-value')).not.toHaveTextContent('21')
   })
 
+  test('handles non-element nodes', () => {
+    const {container} = render(`<span>example</span>`)
+
+    expect(container.querySelector('span').firstChild).toHaveTextContent(
+      'example',
+    )
+  })
+
   test('handles negative test cases', () => {
     const {queryByTestId} = render(`<span data-testid="count-value">2</span>`)
 


nickmccurdy:
This needs more work on the error handling.

eps1lon (commented):
Do we not test the error message? Shouldn't the error message change and mention Node instead of HTMLElement?

nickmccurdy:
You're right. The error message is fixed, just need to fix the rest and add the error message to tests.

gnapse:
#312 reveals that this may not be the only custom matcher that has to allow any node and not HTML elements. It may be worth checking all our matchers to see which other ones make sense to have this change done. Although a quick look at our list of custom matchers makes me think these may be the only two exceptions, but I may be wrong.

nickmccurdy:
That's an issue with DOM Testing Library, not Jest DOM. However we may still want to do that for consistency and to prefer other issues with Jest DOM's types.

src/to-have-text-content.js
```diff
@@ -1,11 +1,15 @@
-import {checkHtmlElement, getMessage, matches, normalize} from './utils'
+import {getMessage, HtmlElementTypeError, matches, normalize} from './utils'
 
 export function toHaveTextContent(
   htmlElement,
   checkWith,
   options = {normalizeWhitespace: true},
 ) {
-  checkHtmlElement(htmlElement, toHaveTextContent, this)
+  const window = htmlElement.ownerDocument.defaultView
```
gnapse, line 8: There should be a check prior to this line. Otherwise this line can throw when receiving something that does not even have a `ownerDocument` property. Or if the obtained `window` value is not a window object and does not have a `window.Node` thing in it.

gnapse (changes requested):
@nickmccurdy I rebased this and tried to work out the CI build failure. It is due to there not being a test that covers the failure case (a test that explicitly makes it fail when it receives something other than a Node).

I then tried to do fix it myself but it became non-trivial. See my comment below. I then decided to leave it to you to decide how to approach this, as you are the PR author. Let me know and can take it from here if you are not able to deal with this in the short term.

julienw:
hey @nickmccurdy, any luck with moving this forward? :-) Needs a hand?

gnapse:
Superseded by #358 

Anyway, thanks @nickmccurdy for giving it a shot here.

PR #310, opened 2020-11-25
ci: add validate workflow based on yo kcd-oss

MatanBobi:
**What**:

Add validate workflow by using `yo kcd-oss`

**Why**:

Because travis no more

**How**:

using `yo kcd-oss`

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged
      

I guess we can reduce the number of node versions in the metrics, but I kept it as is for now..


MatanBobi:
Not sure why this didn't trigger the action 🤔

kentcdodds:
> Not sure why this didn't trigger the action 🤔

Because actions aren't run when the PR comes from a fork. I think you could push this branch to the origin repo and re-open the PR with that branch and it should run.

PR #317, opened 2020-12-17
fix: Ignore comment nodes in toBeEmptyDOMElement

marcelbarner:
**What**:
Adding a logic to the existing `toBeEmptyDOMElement` to ignore comments in the element.

**Why**:
In some frameworks, like Angular, there is a logic to create elements on specific conditions. If this condition is not solved than the framework also adds some comments, this comment has the effect that the matcher `toBeEmptyDOMElement` fails, but the element is empty from the perspective of the user.

**How**:
The matcher checks if the element contains child elements excluded comment nodes.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged


nickmccurdy:
It sounds like a good idea to support comments, but I think it would be easier to just patch `toBeEmpty`. I don't think there needs to be a distinction between empty elements and elements with only comments, as either way the element has no visible content.

marcelbarner:
> It sounds like a good idea to support comments, but I think it would be easier to just patch `toBeEmpty`. I don't think there needs to be a distinction between empty elements and elements with only comments, as either way the element has no visible content.

I am agreed with your point, but than it is not a empty element. An alternative solution will be to combine both logics to a new matcher with the name `toContainNoVisibleContent` or something similar.

nickmccurdy:
It may not technically be empty, but practically speaking there's nothing that would be displayed to the user, and it's better for tests to represent the user's perspective than implementation details.

marcelbarner:
If we look from the user perspective, than I completely agree with your points. I will make the required adjustments.

marcelbarner:
@nickmccurdy should I also add this logic to the matcher `toBeEmpty`? This matcher is marked as deprecated.
Also I have the problem to write a good test to validate that multiline comments are ignored. Can you maybe help me?

gnapse (changes requested):
I had some questions about the regex being used, but then it hit me that I have objections with using regex in the first place. It will most likely be error prone to use a regex approach for this purpose.

I'd suggest we can use the DOM structure itself, instead of parsing anything ourselves. We can traverse the list of nodes inside the target element, and check if it has any child nodes that are not comments.

Here's a snippet that may serve as inspiration:

```js
const nonCommentChildNodes = [...element.childNodes].filter(
    node => node.nodeType !== Node.COMMENT_NODE
)
```

If this list turns out empty, the target element is empty, and vice-versa.

Would love your thoughts on this one.

gnapse:
About this

> should I also add this logic to the matcher toBeEmpty? This matcher is marked as deprecated.

I would not bother adding this fix to the deprecated matcher.

> Also I have the problem to write a good test to validate that multiline comments are ignored. Can you maybe help me?

Using the DOM structure instead of parsing it ourselves will solve this.

marcelbarner:
> I had some questions about the regex being used, but then it hit me that I have objections with using regex in the first place. It will most likely be error prone to use a regex approach for this purpose.
> 
> I'd suggest we can use the DOM structure itself, instead of parsing anything ourselves. We can traverse the list of nodes inside the target element, and check if it has any child nodes that are not comments.
> 
> Here's a snippet that may serve as inspiration:
> 
> ```js
> const nonCommentChildNodes = [...element.childNodes].filter(
>     node => node.nodeType !== Node.COMMENT_NODE
> )
> ```
> 
> If this list turns out empty, the target element is empty, and vice-versa.
> 
> Would love your thoughts on this one.

I think this idea is the better solution to use the DOM structure with that the code will be more stable. I had some concerns about node elements which only contain whitespaces or text, but I tried it and seems to work. I will adjust my implementation.

Thanks for your nice catch.

nickmccurdy:
My mistake, I forgot about the deprecation, I meant `toBeEmptyDOMElement`.

marcelbarner:
No problem I am thank full for all of your great input, so thanks to you all.

marcelbarner:
Now I need your help again, because I am not an expert for `jsdom`.
If I use this this snipped to identify that an node element is a `COMMENT_NODE`
```js
const nonCommentChildNodes = [...element.childNodes].filter(node => node.nodeType !== Node.COMMENT_NODE);
```
I get the following error:
- ReferenceError: Node is not defined

After a research I found the following [issue](https://github.com/jsdom/jsdom/issues/2220). I can not exactly identify what i have to do. I figured out that a workaround is to replace `Node.COMMENT_NODE` with 8 will work but I am not happy with a magic number or to redefine a global definition. So maybe you have an idea what i can do?

gnapse:
I suggest you declare a constant of our own and add a jsdoc comment to it explaining why we can't use `Node.COMMENT_NODE` (including a link to that github issue).

nickmccurdy:
Aren't we targeting the JSDOM environment of Jest which includes these values in the V8 context? They're not really globals, but any code in Jest DOM should be able to pretend they're globals as far as I understand.

README.md
```diff
@@ -215,17 +215,21 @@ expect(getByTestId('not-empty')).not.toBeEmpty()
 toBeEmptyDOMElement()
 ```
 
-This allows you to assert whether an element has content or not.
+This allows you to assert whether an element has no visible content for the user.
+It ignores comments but will fail if the element contains whitespaces.
```
gnapse, line 219:
```suggestion
It ignores comments but will fail if the element contains white-space.
```

src/__tests__/to-be-empty-dom-element.js
```diff
@@ -5,17 +5,33 @@ test('.toBeEmptyDOMElement', () => {
...
         <span data-testid="empty"></span>
         <svg data-testid="svg-empty"></svg>
-    </span>`)
+    </span>
+    <span data-testid="with-comment"><!-- This Comment --></span>
+    <span data-testid="with-multiple-comments"><!-- Comment1 --><!-- Comment2 --></span>
+    <span data-testid="with-element"><span></span></span>
+    <span data-testid="with-element-and-comment"><!--Comment--><span></span></span>
+    <span data-testid="with-whitespace"> </span>
+    <span data-testid="with-whitespace">Text</span>`)
```
gnapse, line 14: I guess the second one should not have the same test id, right? Also, this second element is probably not being used by the test at all.

src/to-be-empty-dom-element.js
```diff
@@ -19,3 +19,16 @@ export function toBeEmptyDOMElement(element) {
...
+ * ℹ Node.COMMENT_NODE can't be used because of the following issue 
+ * https://github.com/jsdom/jsdom/issues/2220
+ *
+ * @param {*} element an HtmlElement or SVGElement
+ * @return {*} true if the element only contains comments or none
+ */
+function isEmptyElement(element){
+  const nonCommentChildNodes = [...element.childNodes].filter(node => node.nodeType !== 8);
+  return nonCommentChildNodes.length === 0;
+}
```
gnapse, line 34: I wonder why there's no new line at the end of this file. If I recall correctly, prettier should be doing it when you commit. But maybe I am mistaken.

src/to-be-empty-dom-element.js
```diff
@@ -19,3 +19,16 @@ export function toBeEmptyDOMElement(element) {
     },
   }
 }
+
+/**
+ * Identifies if a element doesn't contain child elements excluded comments
```
gnapse, line 24:
```suggestion
 * Identifies if an element doesn't contain child nodes (excluding comments)
```

gnapse (changes requested):
Looking good. Requesting changes because of a few relatively minor observations. Let me know what you think.

src/to-be-empty-dom-element.js
marcelbarner: I think you are right, but I have problems to commit with the configured pre-commit hook. It always tries to execute a script which can not be found. The simple reason is that the path doesn't contain any slashes.
At the moment I didn't found any solution only to skip the pre-commit hook, I know this is bad, but better than no commit.
If you know a solution I will be really happy.

---------------------------------------------
Operation System: Windows 10
Script File: ~/node_modules/.bin/kcd-scripts.CMD

gnapse: We probably are not properly setup for a smooth dev experience on Windows.

gnapse (approved):

gnapse:
@all-contributors please add @marcelbarner for code and test

gnapse:
@all-contributors please add @marcelbarner for code and test

PR #318, opened 2020-12-19
fix: Adds expected text for toContainHTML checks that fail (#299)

### LINKED ISSUE #299: toContainHTML no explanation for error message
After receiving HTML from a function, toContainHTML will throw an error without any explanation. No 'expected', terminal will only show what it received. 

Jest test: 
`expect(getElement('paragraph', 'footer')).toContainHTML('<p class="paragraph second-paragraph" id="id7">this is the second paragraph of footer</p>');`

Terminal log: 
` Received:
      <p class="paragraph second-paragraph" id="id7">this is the second paragraph of footer</p>`

If I use toContainElement, it will throw an error that it receives a string and not an HTML element. 

The getElement function will return the HTML element as soon as they find a match with the classes ("class="paragraph second-paragraph")

kiranjd:
**What**:
No expected message when tests fail when `toContainHTML` matcher is used

**Why**:
Fixes #299 

**How**:
Added an expected block with `htmlText` as the expected string with 'Expected' colour using `utils.EXPECTED_COLOR`

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/to-contain-html.js
+++ b/src/__tests__/to-contain-html.js
@@ -2,7 +2,8 @@ import {render} from './helpers/test-utils'
 
 /* eslint-disable max-statements */
 describe('.toContainHTML', () => {
-  const {queryByTestId} = render(`
+  test('handles positive and negative cases', () => {
+    const {queryByTestId} = render(`
     <span data-testid="grandparent">
       <span data-testid="parent">
         <span data-testid="child"></span>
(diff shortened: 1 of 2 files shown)


nickmccurdy:
Can you add a regression test for this please?

kiranjd:
@nickmccurdy  Definitely. Can you point me where I can find a similar test added? Do not know what a regression test is.

kiranjd:
@nickmccurdy added the regression tests

src/__tests__/to-contain-html.js
```diff
@@ -1,7 +1,7 @@
 import {render} from './helpers/test-utils'
 
 /* eslint-disable max-statements */
-test('.toContainHTML', () => {
+describe('.toContainHTML', () => {
```
gnapse, line 4: This change makes no sense. Code inside `describe` blocks is not run as a test. You are effectively disabling all our past tests here.

What you need to do is to either wrap all the previously existing tests inside a new `test` or `it` block inside the `describe`. This block will stand alongside the new `test` block you added.

gnapse (changes requested):

src/__tests__/to-contain-html.js
kiranjd: Read and understood the differences between `describe`, `test` and `it`. Made the changes

nickmccurdy (approved):

gnapse (approved):

PR #319, opened 2020-12-21
fix: check equality in toHaveDisplayValue (fix #290)

### LINKED ISSUE #290: toHaveDisplayValue checks for substring rather than equality when passed string value?
- `@testing-library/jest-dom` version: 5.10.0

### Relevant code or config:

```js
  render(<select data-testid="displayvalue">
    <option value="null"></option>
    <option value="choice" selected>Choice 1</option>
    </select>
    );
  expect(screen.getByTestId("displayvalue")).toHaveDisplayValue(""); // Passes?
  expect(screen.getByTestId("displayvalue")).toHaveDisplayValue(/^$/); // Fails
```

Hi, I'm not sure if this is a bug, but it was pretty surprising behavior to me, and took a while to figure out.  It seems like the toHaveDisplayValue assertion doesn't do an exact match if you pass a string as the expected value, but an "includes" match.  If you want to check that the display value of an element is empty, for example, you have to use a regular expression, as shown above.  (The test is designed to fail)

It also looks like before regular expressions were supported in `.toHaveDisplayValue`, it did do an exact match (though I have not tested this, only looked at the code changes).  It seems to me more intuitive that when passed a string, it should check for a full exact match, especially since that's how `.toHaveValue` works.  Thoughts?  Thank you!

### Example code in a sandbox:

https://codesandbox.io/s/react-testing-library-demo-forked-g5udj?file=/src/__tests__/hello.js

nickmccurdy:
**What**: Uses an equality check on strings passed to `toHaveDisplayValue`

**Why**: #290

**How**: Use `this.equals` (equality check) instead of `matches` (which uses `includes` for substrings)

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/to-have-display-value.js
+++ b/src/__tests__/to-have-display-value.js
@@ -11,6 +11,7 @@ test('it should work as expected', () => {
   `)
 
   expect(queryByTestId('select')).toHaveDisplayValue('Select a fruit...')
+  expect(queryByTestId('select')).not.toHaveDisplayValue('Select')
   expect(queryByTestId('select')).not.toHaveDisplayValue('Banana')
   expect(() =>
     expect(queryByTestId('select')).not.toHaveDisplayValue('Select a fruit...'),
--- a/src/__tests__/to-have-display-value.js
+++ b/src/__tests__/to-have-display-value.js
@@ -152,3 +152,13 @@ test('it should throw if element is not valid', () => {
     `".toHaveDisplayValue() currently does not support input[type=\\"checkbox\\"], try with another matcher instead."`,
   )
 })
+
+test('it should work with numbers', () => {
+  const {queryByTestId} = render(`
+    <select data-testid="select">
+      <option value="">1</option>
+    </select>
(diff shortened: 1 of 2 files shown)


gnapse (approved):
@nickmccurdy technically this is a breaking change, although whoever was relying on this can be considered as having been relying on a bug, so I'm not sure.

nickmccurdy:
It seems like the existing behavior is inaccurate according to the documentation. As users are supposed to follow the documented API, I would consider this to be a semver minor/patch bug fix, not a major release.

nickmccurdy:
@all-contributors add @nickmccurdy for docs, bugs

nickmccurdy:
@all-contributors add @nickmccurdy for docs, code

nickmccurdy:
@all-contributors remove @nickmccurdy for bugs

nickmccurdy:
@all-contributors add @nickmccurdy for docs, code

PR #326, opened 2021-01-07
docs: Update usage in README to make it more copy-pasty

obedparla:
**What**:

Docs small update.

**Why**:

I spent some time going through the docs and the issues to enable it all. Could save people some time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,19 @@ file][]) and you're good to go:
   https://jestjs.io/docs/en/configuration.html#setupfilesafterenv-array
 
 ```javascript
+// In your own jest-setup.ts (or any other name) 
 import '@testing-library/jest-dom'
+
+// jest.config.js add:
+setupFilesAfterEnv: ["<rootDir>/jest-setup.ts"]
+
+// Include your jest-setup.ts if you havent already
+// tsconfig.json
+  "include": [
+    ...
+    "./jest-setup.ts"
+  ],
+
 ```
 
 > Note: If you're using TypeScript, make sure your setup file is a `.ts` and not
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -116,7 +116,19 @@ file][]) and you're good to go:
   https://jestjs.io/docs/en/configuration.html#setupfilesafterenv-array
 
 ```javascript
+// In your own jest-setup.ts (or any other name) 
 import '@testing-library/jest-dom'
+
+// jest.config.js add:
```
gnapse, line 122: nit pick

```suggestion
// In jest.config.js add (if you haven't already)
```

README.md
```diff
@@ -116,7 +116,19 @@ file][]) and you're good to go:
...
+// Include your jest-setup.ts if you havent already
+// tsconfig.json
+  "include": [
+    ...
+    "./jest-setup.ts"
+  ],
+
 ```
 
 > Note: If you're using TypeScript, make sure your setup file is a `.ts` and not
```
gnapse, line 134: The changes in this PR so far assume usage of TypeScript. Maybe leave it all js-centric, and add a sub-header down at this point where we summarize the parts related to TS (different extension, and the snippet to add to tsconfig.json, etc.)

gnapse (commented):
It looks good, but I'm a bit on the fence. We already linked to the jest documentation about the file where jest-dom is to be imported.

Anyway, this is not to object. But I would indeed like to better address the TS vs. JS differences (see comment below). Or hear your thoughts about my thinking.

README.md
obedparla: You're totally right. I've had my head full of TS and forgot about JS lol.

obedparla:
I was able to set it up with the current readme, but I needed to read into the Jest docs and find how to fix TS in the issues here, and I've used Jest a bit. So I figure some people may benefit from more hand-holding.

gnapse (approved):

gnapse:
@all-contributors please add @obedparla for docs

Issue #338, opened 2021-02-05
toBeVisible should imply toBeInTheDocument


calebeby:
### Describe the feature you'd like:

I think that it would make sense for `toBeVisible` to perform, in addition to its CSS-based checks, checks that confirm that the element is in the document.

```js
test('toBeVisible with removed element', () => {
  document.body.innerHTML = "<div id='el'>content</div>"
  const el = document.getElementById('el')
  expect(el).toBeVisible()
  el.remove()
  expect(el).not.toBeVisible() // this line fails, element is "visible"
})
```

If we think about it from a CSS standpoint, yes, there is no CSS that is preventing the element from being visible. But from the user's perspective, the element is definitely not visible. For this reason I think it makes sense to change the behavior of `.toBeVisible` to include the checks from `.toBeInTheDocument`.

gnapse:
Makes sense. Are you up to contribute this fix? If not I may get to it next week.

calebeby:
Yep, I can make a PR. One other thing I have a question about:

`.toBeInTheDocument` has a special case for `null` values:

```js
expect(null).not.toBeInTheDocument() // passes
```

Should `.toBeVisible` also have that case?

```js
expect(null).not.toBeVisible()
```

gnapse:
`toBeInTheDocument` has that special case because a common use is to use a `queryBy*` or even `document.querySelector` and if they fail to return what you queried for, `.toBeInTheDocument`'s most expected behavior would be to indeed let it pass, because that means the queried element was not found.

However, I'd say `toBeVisible` should not do the same. Can you check if it is letting it pass today? I believe it isn't, and in that case even more so, it should not change its functionality in that way.

calebeby:
Yeah right now `expect(null).not.toBeVisible()` does not pass.

gbirke:
Patch #339 broke some Vue tests for me. The tests look like this:

```javascript
const wrapper = mount( SomeComponent, {localVue} );
let myInput = wrapper.find( 'input#foo' );
expect( myInput.element ).toBeVisible();
```

With version 5.11.10 I get "Received element is not visible (element is not in the document)" errors. 

Using Vue 2.6.11 with @vue/test-utils 1.1.3

gnapse:
@gbirke what's `mount` and `wrapper` in your example above?

[...]

PR #339, opened 2021-02-05
`toBeVisible` implies `toBeInTheDocument`

calebeby:
**What/Why**:

https://github.com/testing-library/jest-dom/issues/338

**Checklist**:

- [x] Documentation
- [x] Tests
- [ ] Updated Type Definitions - The `.d.ts` is not in this repo
- [x] Ready to be merged

cc @gnapse


README.md
```diff
@@ -429,6 +430,7 @@ An element is visible if **all** the following conditions are met:
   [`hidden`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden)
   attribute
 - if `<details />` it has the `open` attribute
+- it is present in the document
```
gnapse, line 433: nit pick: I'd put this one first.

src/__tests__/to-be-visible.js
```diff
@@ -242,4 +243,11 @@ describe('.toBeVisible', () => {
       })
     })
   })
+  describe('element not in document', () => {
```
gnapse, line 246: a couple of nit-picks:

```suggestion
  })

  describe('when element not in document', () => {
```

Although my real recommendation here would be to put this test case more towards the beginning, as a sibling case of the first `it` block. I feel there's no need for this to be inside a `describe` block of its own.

gnapse (approved):
Thanks for this!

Left a couple of nit-picky comments. Let me know if/when you address them and we're good to go.

calebeby:
@gnapse Should be all set!

gnapse (approved):

gnapse:
@all-contributors please add @calebeby for code, test

PR #345, opened 2021-03-02
support batch check toBeInTheDocument

bochen2014:
to simplify the check syntax, so instead of 
```js
  expect(htmlElement).toBeInTheDocument()
  expect(svgElement).toBeInTheDocument()
  expect(customElementChild).toBeInTheDocument()
```
you can write
```js
  expect([htmlElement, svgElement, customElementChild]).toBeInTheDocument()

```

### DIFF
--- a/src/__tests__/to-be-in-the-document.js
+++ b/src/__tests__/to-be-in-the-document.js
@@ -33,8 +33,13 @@ test('.toBeInTheDocument', () => {
   expect(htmlElement).toBeInTheDocument()
   expect(svgElement).toBeInTheDocument()
   expect(customElementChild).toBeInTheDocument()
+
+  expect([htmlElement, svgElement, customElementChild]).toBeInTheDocument()
+
   expect(detachedElement).not.toBeInTheDocument()
   expect(nullElement).not.toBeInTheDocument()
+
+  expect([detachedElement, nullElement]).not.toBeInTheDocument()
 
   // negative test cases wrapped in throwError assertions for coverage.
   const expectToBe = /expect.*\.toBeInTheDocument/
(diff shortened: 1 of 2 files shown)


eps1lon (commented):
Thanks for the work.

I don't think we should go down this route. In the end: Why not implement this for all matchers? This doesn't look like it actually improves readability of your test.

Secondly: An important aspect of matchers is their debugging experience. If I pass `[elementA, elementB, elementC]` and `elementB` is not in the document, how does the failure message look like?

gnapse:
Thanks for the interest in improving the library @bochen2014 

Although I have to say I agree with @eps1lon's response. Especially the second point is unarguably a problem. This is a syntactic sugar that will not help with understanding why a test failed when it fails in a matcher using this feature. IMO the benefits are not worth the downsides.

I recommend that new ideas should be proposed first, before implementing them. It may even happen that in other cases instead of rejecting them, we accept it, but even the discussion with others may help refine them. But in this case I'm afraid we shouldn't go down this road.

bochen2014:
thanks @eps1lon and @gnapse for your comments. 

I actually implemented this feature first in my own project (in `setupTest.js`) and then thought it might help others too. 
I have quite a few elements to check in the unit test, so the check code is quite verbose, though to add a syntax sugar to make it cleaner

regarding the concerns
1. yep, this syntax sugar doesn't belong to any particular matcher but rather should be applied to all matchers. I'm only doing it for `to-be-in-the-document` as it's the only matcher used in my project. To extend it to all matchers should be straightforward afaik.
2.  there are 4 scenarios involved with the new syntax (which have been covered in unit test)

```js
expect([exist, exist]).toBeInTheDocument()
expect([nonExist, nonExist]).not.toBeInTheDocument()
expect([exist, nonExist, exist]).toBeInTheDocument() 
expect([nonExist, exist, nonExist]).not.toBeInTheDocument()
```
in above code snippet, case 1 and 2 are happy path, case 4 would name the unexpected found node in error message, the only problems is with case 3 where it would  say `element could not be found in the document` without specifying which.   
There is no easy way to denote that nonExist node and developers would have to use the classic convention to check them one by one. (error message can be changed to `one of the elements could not be found in the document`)

anyway, it's just a syntax sugar, it's not changing the underlying check logic. happy to leave the code as is. thanks again

gnapse:
> one of the elements could not be found in the document

Indeed this is the crux of the issue. If we cannot give a helpful reason for the failure for all situations, then it's not worth it. We'd been trading a slight convenience in syntax for a potential downside in understanding why our tests fail.

Thanks for the understanding. I'll go ahead and close this for now.

PR #349, opened 2021-03-18
fix: normalize expected value in toContainHTML

### LINKED ISSUE #347: toContainHTML does not handle self-closing tags
- `@testing-library/jest-dom` version: 5.11.9 (latest)
- `node` version: n/a
- `npm` (or `yarn`) version: n/a

### Relevant code or config:

```js
import "@testing-library/jest-dom/extend-expect";
import React from "react";
import { render } from "@testing-library/react";

test("demo test", () => {
  const { container } = render(<div className="empty" />);
  // this assertion passes
  expect(container).toContainHTML('<div class="empty"></div>');
  // this one fails, even though it references the same structure
  // expect(container).toContainHTML('<div class="empty" />');  
});
```

### Reproduction:

https://codesandbox.io/s/contains-html-assertion-demo-7nd5n?file=/src/__tests__/hello.js:0-343

### Suggested solution:

toContainHTML matcher treats expected content as a string. I figured that this normalization fixes the matching:

```js
function normalize(content) {
   const div = document.createElement('div');
   div.innerHTML = content;
   return div.innerHTML;
}

expect(container).toContainHTML(normalize('<div class="empty" />'));  // passes  
```

Perhaps the matcher inside could apply this normalization

just-boris:
**What**:

resolve #347

**Why**:

Because this is worth fixing

**How**:

Added `getNormalizedHtml` function, that passes the content to the HTML parser.

This caused two changes in the test case with invalid HTML: `<span data-testid="child"></div>` becomes `<span data-testid="child"></span>`

**Checklist**:

- [ ] Documentation n/a
- [x] Tests
- [ ] Updated Type Definitions n/a
- [ ] Ready to be merged

### DIFF
--- a/src/__tests__/to-contain-html.js
+++ b/src/__tests__/to-contain-html.js
@@ -18,6 +18,7 @@ describe('.toContainHTML', () => {
     const nonExistantElement = queryByTestId('not-exists')
     const fakeElement = {thisIsNot: 'an html element'}
     const stringChildElement = '<span data-testid="child"></span>'
+    const stringChildElementSelfClosing = '<span data-testid="child" />'
     const incorrectStringHtml = '<span data-testid="child"></div>'
     const nonExistantString = '<span> Does not exists </span>'
     const svgElement = queryByTestId('svg-element')
@@ -25,14 +26,14 @@ describe('.toContainHTML', () => {
     expect(grandparent).toContainHTML(stringChildElement)
     expect(parent).toContainHTML(stringChildElement)
     expect(child).toContainHTML(stringChildElement)
+    expect(child).toContainHTML(stringChildElementSelfClosing)
     expect(grandparent).not.toContainHTML(nonExistantString)
     expect(parent).not.toContainHTML(nonExistantString)
     expect(child).not.toContainHTML(nonExistantString)
     expect(child).not.toContainHTML(nonExistantString)
-    expect(grandparent).not.toContainHTML(incorrectStringHtml)
-    expect(parent).not.toContainHTML(incorrectStringHtml)
-    expect(child).not.toContainHTML(incorrectStringHtml)
-    expect(child).not.toContainHTML(incorrectStringHtml)
+    expect(grandparent).toContainHTML(incorrectStringHtml)
+    expect(parent).toContainHTML(incorrectStringHtml)
+    expect(child).toContainHTML(incorrectStringHtml)
 
     // negative test cases wrapped in throwError assertions for coverage.
     expect(() =>
@@ -59,6 +60,9 @@ describe('.toContainHTML', () => {
     expect(() =>
       expect(child).not.toContainHTML(stringChildElement),
     ).toThrowError()
+    expect(() =>
+      expect(child).not.toContainHTML(stringChildElementSelfClosing),
+    ).toThrowError()
     expect(() => expect(child).toContainHTML(nonExistantString)).toThrowError()
     expect(() => expect(parent).toContainHTML(nonExistantString)).toThrowError()
     expect(() =>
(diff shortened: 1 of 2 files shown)


src/__tests__/to-contain-html.js
```diff
@@ -18,21 +18,23 @@ describe('.toContainHTML', () => {
     const nonExistantElement = queryByTestId('not-exists')
     const fakeElement = {thisIsNot: 'an html element'}
     const stringChildElement = '<span data-testid="child"></span>'
-    const incorrectStringHtml = '<span data-testid="child"></div>'
+    const stringChildElementSelfClosing = '<span data-testid="child" />'
+    // const incorrectStringHtml = '<span data-testid="child"></div>'
```
just-boris, line 22: The new change also normalizes this invalid HTML, so this also now passes:

```js
expect(child).toContainHTML('<span data-testid="child"></div>');  // normalized to `<span data-testid="child"></span>`
```

src/to-contain-html.js
```diff
@@ -1,10 +1,20 @@
...
+  div.innerHTML = htmlText
+  return div.innerHTML
+}
+
 export function toContainHTML(container, htmlText) {
   checkHtmlElement(container, toContainHTML, this)
 
+  if (typeof htmlText !== 'string') {
+    throw new Error(`.toContainHTML() expects a string value, got ${htmlText}`)
+  }
```
just-boris, line 14: Without this change, the `nonExistantElement` tests started failing. This value is basically an equivalent of `expect(element).toContainHTML(null)` which turned into a `string.includes(null)`, producing `false`. 

Now the `htmlText` always gets converted to a string, so this extra check is needed to keep the assertions failing where they were before.

just-boris:
Opened this as a draft, as there are two maybe-major changes, see the comments above

gnapse:
@just-boris can you take this PR to a valid non-draft state (e.g. uncomment or remove the commented code) so that I can review. I cannot get a full idea of what's going on or what you are proposing without knowing what about all that commented out code.

just-boris:
@gnapse 
Updated the PR. It introduces two notable changes besides fixing the https://github.com/testing-library/jest-dom/issues/347

1. non-string values throw a different error now. It was an assertion error and now this is precise `.toContainHTML() expects a string value, got ${htmlText}`
2. Invalid HTML gets normalised now, passing the assertion if the result is the same as the actual content

src/__tests__/to-contain-html.js
```diff
@@ -18,21 +18,18 @@ describe('.toContainHTML', () => {
...
     expect(child).toContainHTML(stringChildElement)
+    expect(child).toContainHTML(stringChildElementSelfClosing)
     expect(grandparent).not.toContainHTML(nonExistantString)
     expect(parent).not.toContainHTML(nonExistantString)
     expect(child).not.toContainHTML(nonExistantString)
     expect(child).not.toContainHTML(nonExistantString)
-    expect(grandparent).not.toContainHTML(incorrectStringHtml)
-    expect(parent).not.toContainHTML(incorrectStringHtml)
-    expect(child).not.toContainHTML(incorrectStringHtml)
-    expect(child).not.toContainHTML(incorrectStringHtml)
```
gnapse, line 35: Do we need to remove the test on the invalid html? What happens after these changes if we give invalid html like the one in this string? It'd be nice to know and add a test for that. And hopefully that what it does is still close to what one would expect.

gnapse (commented):
I'm mostly OK with these changes. I left a comment on the only thing that worries me. Would like to hear what you think before merging this.

src/__tests__/to-contain-html.js
just-boris: I agree, the tests can stay, just need to flip the direction, as this assertion now passes.

Fixed in the next revision

gnapse: How is it that these tests now pass? I would expect them not to pass. The rendered DOM in this test does contain the invalid html given. 🤔

just-boris: HTML parser ignores unmatching closing tags. Currently we have this content: `<span data-testid="child"></div>`. After normalization via `innerHTML` it becomes `<span data-testid="child"></span>` and this matches the actual content.

With normalization, this test will not pass, for example:

```js
const {container} = render(`<table>
  <tr>
    <td>test</td>
  </tr>
</table>`)
expect(container).toContainHTML('<table>test</table>');
```

`<table>test</table>` does not expand to the full table and does not match the real content

gnapse: Hmmm, I wonder if we could detect that, and disallow invalid html. This also makes it technically a breaking change, although I'm thinking that I'll pass on that, since it is very unlikely someone was passing invalid html to explicitly expect a failure.

I think this is good to go. Will merge soon.

gnapse (approved):

src/__tests__/to-contain-html.js
just-boris: I am not sure this is possible without rolling your own parser. Built-in DOM parser does not expose enough information.

I am also wondering about the use-case here. Spotting the typos is nice, but is there any more important case where it is needed?

gnapse: Not really, which is why I said I approve anyway. It's just a bit unintuitive, but I agree that it makes sense since it is how html works anyway. Developing our own parser is out of the question. This matcher is anyway an outlier, in the sense that we do not recommed its use except in some unique circumstances.

gnapse:
@all-contributors please add @just-boris for code, test, bug

Issue #355, opened 2021-04-19
Update README to correct Usage section


kylejw2:
<!--

I was following the `README.md` and wrote a simple test:
```typescript
import React from 'react';
import PrimaryButton from './Primary';
import { render, screen } from '@testing-library/react';

describe('basic test', () => {
  it('renders the children passed to it', async () => {
    render(<PrimaryButton>Hello</PrimaryButton>);
    const HelloButton = await screen.findByText('Hello');
    expect(HelloButton).toBeInTheDocument();
  });
});
```

I was continually getting an error, saying `Property 'toBeInTheDocument' does not exist on type 'JestMatchers<HTMLElement>'`. I followed the `README.md` exactly as it's written so I assumed everything would work. In my `jest-setup.ts` file I had `import '@testing-library/jest-dom';` but the correct implementation was `import '@testing-library/jest-dom/extend-expect';`

The solution has been found, so I'm requesting that you update the documentation as it's currently missing this piece of information.

Thanks.

PR #356, opened 2021-04-20
Update missing documentation in README

kylejw2:
This PR is linked to this issue I created https://github.com/testing-library/jest-dom/issues/355.

When following the documentation before these updates, there's a strong change the user will encounter an error because the `extend-expect` module hasn't been imported. This PR includes the missing documentation and should help users avoid this error or easily troubleshoot it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ file][]) and you're good to go:
 ```javascript
 // In your own jest-setup.js (or any other name)
 import '@testing-library/jest-dom'
-// Can include the extend-expect extension if needing to test components against the DOM
+// Can also include the extend-expect extension in setup file
 import '@testing-library/jest-dom/extend-expect'
 
 // In jest.config.js add (if you haven't already)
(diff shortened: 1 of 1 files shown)


gnapse (changes requested):
Importing the package's root module will already do what `extend-expect` used to do. That is, importing this:

```js
// In your own jest-setup.js (or any other name)
import '@testing-library/jest-dom'
```

is already enough. We haven't removed `extend-expect`, but I was planning to make a new major release soon with this and at least one other breaking change.

Can you comment why you think you need to import this in addition to already being importing `@testing-library/jest-dom`? Maybe I'm missing something.

kylejw2:
I assumed that only having the following would work, given the documentation.
```typescript
// In your own jest-setup.js (or any other name)
import '@testing-library/jest-dom'
```
Correct me if I'm wrong please, but importing `jest-dom` through the above approach only pulls in the `index` file of the `jest-dom` folder. Perhaps it didn't work as expected for me because I couldn't find an `index` file in the `jest-dom` folder.
Screen Shot 2021-04-21 at 10 19 58 AM
Here's the file/folder structure I have from installing the `jest-dom` library. There's no `index` file so I had to directly import `extend-expect` in order to access its methods.

gnapse:
The index file lives inside the `dist` folder, and it is considered the main entry point when importing the package thanks to [this line](https://github.com/testing-library/jest-dom/blob/21ad89bc707b6e41af887d1bd103dd6dbc665dee/package.json#L5) in the `package.json` file.

What version of jest-dom are you using when you get this behaviour of the regular package import not working properly?

kylejw2:
I think you may have located the issue. I'm using version 4.2.4

gnapse:
Yup, that's what I thought. I'm going to close this for now.

PR #358, opened 2021-04-22
feat: check for any Node in toHaveTextContent

### LINKED ISSUE #306: toHaveTextContent should support any Node

julienw:
Fixes #306

I based it off #307 but changed a lot of things, that's why I didn't keep the initial author information.

**What, Why**: Please have a look at #306
**How**: I added a `checkNode` function in `utils` so that it can be reused elsewhere. I followed the style of the other functions and also the comment in #307 about checking that we have a window before using it.

**Checklist**:

- [X] Documentation
- [X] Tests
- [ ] Updated Type Definitions <-- Not sure where to check them?
- [X] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -748,7 +748,8 @@ The usual rules of css precedence apply.
 toHaveTextContent(text: string | RegExp, options?: {normalizeWhitespace: boolean})
 ```
 
-This allows you to check whether the given element has a text content or not.
+This allows you to check whether the given node has a text content or not. This
+supports elements, but also text nodes and fragments.
 
 When a `string` argument is passed through, it will perform a partial
 case-sensitive match to the element content.
--- a/README.md
+++ b/README.md
@@ -752,7 +752,7 @@ This allows you to check whether the given node has a text content or not. This
 supports elements, but also text nodes and fragments.
 
 When a `string` argument is passed through, it will perform a partial
-case-sensitive match to the element content.
+case-sensitive match to the node content.
 
 To perform a case-insensitive match, you can use a `RegExp` with the `/i`
 modifier.
--- a/src/utils.js
+++ b/src/utils.js
@@ -2,8 +2,8 @@ import redent from 'redent'
 import {parse} from 'css'
 import isEqual from 'lodash/isEqual'
 
-class HtmlElementTypeError extends Error {
-  constructor(received, matcherFn, context) {
+class GenericTypeError extends Error {
+  constructor(expectedString, received, matcherFn, context) {
     super()
 
     /* istanbul ignore next */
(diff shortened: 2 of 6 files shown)


src/utils.js
```diff
@@ -31,24 +31,45 @@ class HtmlElementTypeError extends Error {
...
+  }
+}
+
+class NodeTypeError extends GenericTypeError {
+  constructor(...args) {
+    super('be a Node', ...args)
+  }
+}
+
+function checkHasWindow(htmlElement, ErrorClass, ...args) {
```
gnapse, line 52: I think you could've kept the signature of this one intact as we only call it internally in this module, only once, and always to check for the HTML element check, and not the Node check.

Not a big deal or a blocking change request. Can stay as is. Maybe I'm missing something.

gnapse (approved):

gnapse:
> Updated Type Definitions <-- Not sure where to check them?

This will need to be handled separately in https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/testing-library__jest-dom

README.md
```diff
@@ -748,7 +748,8 @@ The usual rules of css precedence apply.
 toHaveTextContent(text: string | RegExp, options?: {normalizeWhitespace: boolean})
 ```
 
-This allows you to check whether the given element has a text content or not.
+This allows you to check whether the given node has a text content or not. This
+supports elements, but also text nodes and fragments.
 
 When a `string` argument is passed through, it will perform a partial
 case-sensitive match to the element content.
```
gnapse, line 755:
```suggestion
case-sensitive match to the node content.
```

src/utils.js
julienw: See line 63, I actually call it for the node check ;-)

gnapse: yes, I know.

My point was that that was the only call to it. But never mind, it's ok as is. It was a bit of a nit-pick, and maybe a wrong one at that anyway.

julienw:
> This will need to be handled separately in https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/testing-library__jest-dom

I see, looking at this I believe there's no change needed. But I'm not using TS (we're using Flow in our project) so I may miss something.

For Flow, what we put in `expect` is typed as `any` so everything passes.

julienw:
Thanks for the quick review and merge!

gnapse:
Thank you too for your contribution!

gnapse:
@all-contributors add @julienw for code, test

src/utils.js
julienw: ah yeah I see what you mean :-) yeah, maybe it's a bit over-designed :-)

nickmccurdy:
The type declarations live in [`@types/testing-library__jest-dom`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/testing-library__jest-dom). However as far as I understand, the type of the expected value isn't actually checked, so I don't think you need to update them in this case.

RandinAk:
> Fixes #306
> 
> 
> 
> I based it off #307 but changed a lot of things, that's why I didn't keep the initial author information.
> 
> 
> 
> 
> 
> 
> 
> **What, Why**: Please have a look at #306
> 
> **How**: I added a `checkNode` function in `utils` so that it can be reused elsewhere. I followed the style of the other functions and also the comment in #307 about checking that we have a window before using it.
> 
> 
> 
> **Checklist**:
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> - [X] Documentation
> 
> - [X] Tests
> 
> - [ ] Updated Type Definitions <-- Not sure where to check them?
> 
> - [X] Ready to be merged
> 
>       
> 
> 
> 
> 
> 
>

gnapse:
@RandinAk what's the purpose of your last comment? It seems to be mostly a quote of this PR's description, with no extra info added.

gnapse:
@all-contributors add @julienw for code, test

PR #363, opened 2021-05-05
doc: Better documentation for toContaintHTML

### LINKED ISSUE #362: Version 5.11.10 broke toContainHTML
- `@testing-library/jest-dom` version: 5.11.10
- `node` version: v10.19.0
- `yarn` version: 1.22.5
- `@testing-library/react` version: 11.2.6

### Relevant code or config:

```js
const { baseElement } = render('<div class="Loader"><div class="Child"></div></div>');
expect(baseElement).toContainHTML('<div class="Loader">');
```

### What you did:

We are testing that we are rendering a loader by checking if a div with class 'Loader' was rendered.
This div has children which we do not have control over.

### What happened:

Previously the test worked fine, but after upgrading to 5.11.10+ the test started failing.

### Reproduction:

Render two nested divs, then check only for opening tag of outer div.
```js
const { baseElement } = render('<div class="Loader"><div class="Child"></div></div>');
expect(baseElement).toContainHTML('<div class="Loader">');
```

### Problem description:

Investigating the code showed that normalizeHtml adds a closing div tag to the string,
then it doesn't match the rendered html anymore because the div has children.
So, it tries to match ``` <div class="Loader"></div> ```.

Error Message:
```
expect(element).toContainHTML()
    Expected:
      <div class="Loader">
    Received:
      <body><div class="Loader"><div class="Child"></div></div></body>
```

### Suggested solution:

Adding the html to the dom (in normalizeHtml) introduces some unexpected behavior.
The printed error message/jest ouput doesn't reflect such behavior, it's also not documented.

gnapse:
**What**:

More explicit requirements in the documentation of `toContainHTML`.

**Why**:

https://github.com/testing-library/jest-dom/issues/362#issuecomment-832100157

**How**:

**Checklist**:

- [x] Documentation
- [x] Ready to be merged

Fixes #362.

### DIFF
--- a/README.md
+++ b/README.md
@@ -491,7 +491,7 @@ toContainHTML(htmlText: string)
 ```
 
 Assert whether a string representing a HTML element is contained in another
-element. The string should contain valid html, and not any substrings of html.
+element. The string should contain valid html, and not any incomplete html.
 
 #### Examples
 
@@ -500,9 +500,15 @@ element. The string should contain valid html, and not any substrings of html.
 ```
 
 ```javascript
+// These are valid uses
 expect(getByTestId('parent')).toContainHTML('<span data-testid="child"></span>')
 expect(getByTestId('parent')).toContainHTML('<span data-testid="child" />')
 expect(getByTestId('parent')).not.toContainHTML('<br />')
+
+// These won't work
+expect(getByTestId('parent')).toContainHTML('data-testid="child"')
+expect(getByTestId('parent')).toContainHTML('data-testid')
+expect(getByTestId('parent')).toContainHTML('</span>')
 ```
 
 > Chances are you probably do not need to use this matcher. We encourage testing
--- a/README.md
+++ b/README.md
@@ -491,7 +491,7 @@ toContainHTML(htmlText: string)
 ```
 
 Assert whether a string representing a HTML element is contained in another
-element:
+element. The string should contain valid html, and not any substrings of html.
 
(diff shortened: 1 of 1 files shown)


gnapse:
I tagged some folks here to get an approving review and a second pair of eyes and your opinion. Thanks!

README.md
```diff
@@ -491,7 +491,7 @@ toContainHTML(htmlText: string)
 ```
 
 Assert whether a string representing a HTML element is contained in another
-element:
+element. The string should contain valid html, and not any substrings of html.
```
eps1lon, line 494: "substrings of html" is a bit unclear to me. It sounds like "<p></p>" would be considered a "substring" but that's valid HTML. An example would help here.

gnapse: I agree this is probably not the clearest terms. But the examples are below. Do you think we need examples of what won't work? Like for instance, this:

```js
// These won't work
render(<div class="test">text</div>)
expect(element).toContainHTML('</div>')
expect(element).toContainHTML('class="test"')
```

eps1lon: I think "partial HTML" or "incomplete HTML" is more fitting here. 

Oh and move the comment since the render does work.
```diff
-// These won't work
render(<div class="test">text</div>)
+// These won't work
expect(element).toContainHTML('</div>')
expect(element).toContainHTML('class="test"')
```

eps1lon (approved):

Issue #364, opened 2021-05-10
`.toHaveValue` expect string values for input type range 


Okazari:
- `@testing-library/jest-dom` version: 5.11.9
- `node` version: 12.16.0
- `npm` (or `yarn`) version: 1.22.5

### Relevant code or config:

```js

const MyComponent = () => <input data-test="input" type="range" value={5} onChange={() => {}} />

  it('should work', () => {
    render(<MyComponent />);
    let inputRange = screen.getByTestId('input');
    expect(inputRange).toHaveValue(5);
  });
```

Hey there ! I just stumble upon a strange case were testing value for a input type range expect values to be string.
I find it quite odd since those input value are always numbers.

### Reproduction:

Here is a minimap repro in a codesandbox https://codesandbox.io/s/react-testing-library-demo-forked-pve4d?file=/src/__tests__/hello.js

### Problem description:

 - `.toHaveValue` expect string values for input type range 

It force the test to use strings instead of number. It is annoying when dealing with const that are numbers.

### Suggested solution:

- `.toHaveValue` ccould expect number values for input type range

Okazari:
I may open a PR for that if you can point me where the job is done for `input type="number"` :+1:

gnapse:
Thanks for your bug report. Indeed, this should work the same as with `<input type="number" />`. Here's where we treat numeric inputs differently: https://github.com/testing-library/jest-dom/blob/637529e3892611092671f30c135587c130fbf92d/src/utils.js#L191-L192

camsteffen:
I think it's unfortunate that there is any transformation. I expect `toHaveValue` to simply match `element.value`. This has actually led me down a bad path where I assumed I had a number type in my code that was actually a string.

We could have `toHaveValueAsNumber` that uses `element.valueAsNumber`.

PR #365, opened 2021-05-12
fix: `.toHaveValue()` / `.toHaveFormValues()` on input type range now expect a number

Okazari:
**What**:

This PR aim to fix https://github.com/testing-library/jest-dom/issues/364.

**Why**:

In my use case on a react app, i'm using a constant that is a number for a default value of an input type range. I couldn't use it in my test without hacking a bit since `.toHaveValue()` were expecting a string instead of a number.

**How**:

It makes jest-dom consider `<input type="range />` value getter the same as `<input type="number" />`

**Checklist**:

- [ ] Documentation
- [X] Tests
- [X] Ready to be merged


Okazari:
Hey there :wave: the PR template mention to `Update type definitions` but i didn't manage to find any in the repo. Could you point me were are those ?

Thanks :smile:

nickmccurdy:
[They're in DefinitelyTyped.](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/testing-library__jest-dom)

nickmccurdy (approved):

nickmccurdy:
This looks good, but should we leave the breaking changes for a major release or consider this a bug fix?

gnapse:
FWIW, I was planning to take some time next week to gather this and at least one other PR that are breaking changes, and make a single new major release with them all. I was also planning to include in this major release the removal of all the custom matchers that we already have marked as deprecated for various versions now. There are 2 or 3 of them.

nickmccurdy:
Sounds like a plan. Let me know if I can help with anything.

PR #377, opened 2021-06-10
feat: toHaveAccessibleName and toHaveAccessibleDescription

### LINKED ISSUE #111: Create new matcher `toBeLabelled`

The idea is to have a custom matcher on the model of `toBeRequired` or `toBeDisabled` helping folks to know if they labelled correctly the markup they are testing.

This helper would be interesting for me because there is a variety of ways to label correctly an element, such as:

* `aria-label`
* `aria-labelledby`
* `alt` on images
* `title` on images and SVGs
*  having a text content (by using `.toHaveTextContent`)

The main limitation is that I am not sure we can easily know when a DOM node is labelled thanks to a nearby _visually-hidden_ or displayed text.

### Suggested implementation:

Just doing all the tests decribed few lines before.

Doing each test one after the other -or- using `jest-axe` but in this case lacking flexibility on doing more granular tests.

* [WCAG 2.0 G94: Providing short text alternative for non-text content that serves the same purpose and presents the same information as the non-text content](https://www.w3.org/TR/WCAG20-TECHS/G94.html)
* https://www.sarasoueidan.com/blog/accessible-icon-buttons/
* https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute
* https://www.w3.org/TR/wai-aria/#aria-label

I could help with the implementation if you're up to adding this to this library :+1:

gnapse:
**What**:

Add two new custom matchers: `toHaveAccessibleName` and `toHaveAccessibleDescription`.

Closes #111.
Supersedes #112.

**Why**:

These enhance the abilities to test a UI with accessibility in mind, and encourage developers to properly label and describe the UI elements in their apps.

**How**:

The heavy lifting of the work here is delegated to [`dom-accessibility-api`](https://github.com/eps1lon/dom-accessibility-api).

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -533,7 +533,7 @@ expect(getByTestId('parent')).toContainHTML('</span>')
 toHaveAccessibleDescription(expectedAccessibleDescription?: string | RegExp)
 ```
 
-This allows to assert that an element is has the expected
+This allows to assert that an element has the expected
 [accessible description](https://w3c.github.io/accname/).
 
 You can pass the exact string of the expected accessible description, or you can


README.md
```diff
@@ -526,6 +527,94 @@ expect(getByTestId('parent')).toContainHTML('</span>')
 
 <hr />
 
+### `toHaveAccessibleDescription`
+
+```typescript
+toHaveAccessibleDescription(expectedAccessibleDescription?: string | RegExp)
+```
+
+This allows to assert that an element is has the expected
```
eps1lon, line 536:
```suggestion
This allows to assert that an element has the expected
```

eps1lon (approved):
Just glanced over it and couldn't spot anything that stands out. Looks good :+1:

SevenOutman:
What’s the difference between `toHaveAccessibleDescription` and the existing `toHaveDescription`?

gnapse:
Ooohh, I knew I was forgetting something. That one has to be deprecated. It covers only the `aria-describedby` attribute. I'll create a new release deprecating it.

SevenOutman:
> Or maybe I can make it an alias 🤔

I think whether to make it an alias is a decision on naming convention. For example the matcher for “to be accessibly valid” is named `toBeValid` rather than “toBeAccessiblyValid”. Another example is the `name` option in `getByRole` from `testing-library/dom`. I guess the logic here is to hide those overelaborate specifications behind a set of easy-to-read API.

On the other hand, we also have some other APIs which are not for testing a11y. Shortcutting a11y matchers may lead to confusion.

gnapse:
Valid points. Sometimes it all goes to make some compromises.

- `toHaveDescription` could've been nice to keep around, but `toHaveName` is certainly too ambiguous, because it could lead folks to believe that it matches the `name` attributes of `input` fields, for instance. These two are certainly and fully about accessibility features, so the extra verbosity on the name is not bad IMO. The alignment in both namings as well. I would hate having `toHaveDescription` alongside `toHaveAccessibeName`. We could've also call it `toBeLabelled` instead of `toHaveAccessibleName`, but again, it does not pair well with the `toHave` prefix of `toHaveDescription`. And also, the accessible name is technically not always a label (e.g. svg title).
- `toBeValid` does not only check the `aria-` attributes, but also the [`checkValidity` API](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Constraint_validation).
- The `name` in `getByRole` is ok because there's a more specific meaning of what an element's name is within the context of speaking of aria roles.

I will not claim that I thought about all this beforehand, either. You've certainly made me think about it more now after the fact.

Would you say you see something off with the direction we've taken in general?

SevenOutman:
`toHaveDescription` is not so ambiguous as `toHaveName` IMO. It's kinda like `toBeChecked` where "be checked" can match several conditions. Naming a `toHaveAccessibleDescription` feels just like doing a `toHaveAccessibleErrorMessage` to me while "error message" is not ambiguous at all.

I'd prefer reserving `toHaveDescription`. It doesn't have to pair with the `toHaveAccessibleName`. Aligning the `toHaveAccessible-` prefix may confuse users with those matchers that don't have this prefix while they actually checks a11y attributes.

gnapse:
Fair enough. FWIW, the naming was also influenced by how this feature is named in the very spec that defines it: https://w3c.github.io/accname/

I agree that `toHaveDescription` could've been kept. I do like the alignment in naming.

PR #414, opened 2021-10-15
Update aria-query to 5.0.0

### LINKED ISSUE #413: Update aria-query to avoid babel runtime

I'd like to suggest an update to aria-query@5.  They removed their usage of Map and Set, and by doing so, removed babel as runtime dependencies.  I'm currently struggling with one of the babel dependencies that is being brought in to my project via this package, and it could save disk space for other consumers who are not using babel as well.

IanVS:
**What**:

This updates the version of `aria-query` to 5.0.0.  

**Why**:

Closes #413.

**How**:

This is a breaking change in that Map and Set are no longer used.  But, the `roles` object that is returned has methods on it, including `.get()`, which is used in `roleSupportsChecked`.

The only update I made to the code is to remove the `Array.from()` call, since `roles.keys()` already returns an array now.

**Checklist**:

- [ ] Documentation N/A
- [X] Tests
- [X] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
   "dependencies": {
     "@babel/runtime": "^7.9.2",
     "@types/testing-library__jest-dom": "^5.9.1",
-    "aria-query": "^4.2.2",
+    "aria-query": "^5.0.0",
     "chalk": "^3.0.0",
     "css": "^3.0.0",
     "css.escape": "^1.5.1",
--- a/src/to-be-checked.js
+++ b/src/to-be-checked.js
@@ -57,7 +57,7 @@ function supportedRolesSentence() {
 }
 
 function supportedRoles() {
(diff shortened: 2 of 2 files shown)


IanVS:
Hi, the corresponding PR was merged into dom-testing-library.  Would anyone like to take a look at this PR?  I think it's best if the versions stay in-sync.

gnapse:
To confirm what you seem to say in the PR description: is this a breaking change to the outside world?

IanVS:
I do not believe this is breaking, and a major release was not needed for dom-testing-library, see: https://github.com/testing-library/dom-testing-library/pull/1058#issuecomment-945597465

gnapse (approved):

gnapse:
@all-contributors please add @IanVS for code

IanVS:
Thanks @gnapse!

PR #415, opened 2021-10-19
import parse directly from css

yannbf:
Hey everyone! Thanks for this incredible library 🌈 

**What**:

At Storybook we are developing an addon that will allow developers to run interaction tests directly in the browser, using testing library. We'd also love to include the matcher extenders so that developers can use the assertions from `jest-dom/extend-expect`, however importing things directly from `css` result in the following failure:
```Module not found: Error: Can't resolve 'fs'```

This is an issue that happened as well in the `testing-library/jasmine-dom` package and was addressed with the exact same fix as this PR: https://github.com/testing-library/jasmine-dom/pull/8

**Why**:

These changes would allow this library to run in the browser without breaking

**How**:

Given that the library only needs `cssParse`, there is no need to import it from the root of the package. Importing it from the root makes other files be imported as well, which results in the error mentioned above.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,5 +1,5 @@
 import redent from 'redent'
-import {parse} from 'css'
+import cssParse from 'css/lib/parse'
 import isEqual from 'lodash/isEqual'
 
 class GenericTypeError extends Error {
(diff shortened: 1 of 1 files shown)


nickmccurdy (approved):
Good thinking, thanks for the contribution!

nickmccurdy:
@all-contributors add @yannbf and @IanGrainger for code

IanGrainger:
Looks Identical to: https://github.com/testing-library/jasmine-dom/pull/8 - which was the change for Jasmine - so seems good! 👍 thanks for the @ :)

PR #421, opened 2021-11-23
[Feature] Support of custom select with role `listbox`

Joandres-Lara:
**What**: Support to role "listbox" element with option children values.

**Why**: Custom select elements is not supported, and somethings custom dropdown elements no using tag `<select>` native.

**How**: We can use a div tag with a `listbox` role, and a series of children, also div, with the` option` role 

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/__tests__/to-have-value.js
+++ b/src/__tests__/to-have-value.js
@@ -82,6 +82,41 @@ describe('.toHaveValue', () => {
     expect(queryByTestId('single')).toHaveValue('first')
   })
 
+  test('handles values of custom select with role `listbox`', () => {
+   const {queryByTestId} = render(`
+     <div role="listbox" data-testid="single">
+       <div role="option" value="first">First Value</div>
+       <div role="option" value="second" aria-selected="true">Second Value</div>
+       <div role="option" value="third">Third Value</d>
+     </div>
+ 
+     <div role="listbox" data-testid="multiple" aria-multiselectable>
+       <div role="option" value="first">First Value</div>
+       <div role="option" value="second" aria-selected="true">Second Value</div>
+       <div role="option" value="third" aria-selected="true">Third Value</div>
+     </div>
+ 
+     <div role="listbox" data-testid="not-selected" >
+       <div role="option" value="" aria-disabled aria-selected="true">- Select some value - </div>
+       <div role="option" value="first">First Value</div>
+       <div role="option" value="second">Second Value</div>
+       <div role="option" value="third">Third Value</div>
+     </div>
+   `)
+ 
+   expect(queryByTestId('single')).toHaveValue('second')
+   expect(queryByTestId('single')).toHaveValue()
+ 
+   expect(queryByTestId('multiple')).toHaveValue(['second', 'third'])
+   expect(queryByTestId('multiple')).toHaveValue()
+ 
+   expect(queryByTestId('not-selected')).not.toHaveValue()
+   expect(queryByTestId('not-selected')).toHaveValue('')
+ 
+   queryByTestId('single').children[0].setAttribute('aria-selected', true)
+   expect(queryByTestId('single')).toHaveValue('first')
+  });
+
   test('handles value of textarea element', () => {
     const {queryByTestId} = render(`
       <textarea data-testid="textarea">text value</textarea>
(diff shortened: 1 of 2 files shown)


src/__tests__/to-have-value.js
```diff
@@ -82,6 +82,41 @@ describe('.toHaveValue', () => {
...
+     <div role="listbox" data-testid="single">
+       <div role="option" value="first">First Value</div>
+       <div role="option" value="second" aria-selected="true">Second Value</div>
+       <div role="option" value="third">Third Value</d>
+     </div>
+ 
+     <div role="listbox" data-testid="multiple" aria-multiselectable>
+       <div role="option" value="first">First Value</div>
+       <div role="option" value="second" aria-selected="true">Second Value</div>
+       <div role="option" value="third" aria-selected="true">Third Value</div>
```
gnapse, line 96: I'm not sure that `value` is a standard attribute of `div` elements. I'm not even sure this pattern of `role="option" value="…"` is documented or valid.

gnapse (commented):
Hi, thanks for your contribution.

I'm concerned about this direction you are proposing, though. It is not clear that this is a valid or documented pattern. jest-dom cannot take a direction that caters to custom approaches that people may be taking in their own apps. It's ok if this works for you, but it is not a standard DOM structure.

The recommended approach for custom elements that aim to be form elements is to include a `<input type="hidden" value="…" />` with the value that the element has. This ensures that it can be transparently handled by a form being submitted, for instance.

Can you point out official HTML/DOM/ARIA documentation that validates this `role="option" value="…"` as something to be recommended and supported?

src/__tests__/to-have-value.js
Joandres-Lara: Hi @gnapse, thank's by feedback.

https://www.w3.org/TR/wai-aria-1.1/#option, this is ARIA documentation oficial, and specified base concept of role "option", is a option element; https://html.spec.whatwg.org/multipage/form-elements.html#the-option-element, I'm aware that the documentation does not explicitly say that you need an attribute named `value`, maybe a good option to use would be` data-value`. 

`<div role="option" data-value="first">First value</div>`

This a example why is necesary, Here is a sandbox, which illustrates the problem, with an input type = "hidden", the access would not be so transparent, and it would be necessary to obtain this to make the assertion of the value. 

https://codesandbox.io/s/bold-lehmann-cwfw0?file=/tests/CustomSelect.test.js

gnapse: I understand your concern and your need. I've worked both implementing custom form-like components, and used them.

However, my concern here is that this is a generic DOM assertions library. The approach you use to define the "value" of each of this custom option elements is specific to your app. One library may be using this `value` attribute (which is strictly invalid) while another library may be using the more acceptable `data-value` to keep this data in the DOM. But this assertion library should not pick sides and implement a custom assertion that assumes one of this app-specific solutions to this problem.

That's why I suggested that this may not be needed at all. What I really think a component like this should do is to provide a `<input type="hidden" value="…" /> that's kept up-to-date with what the current value is.

Joandres-Lara: I understand, if is necessary to modify the pull request, so that suits the needs of the library, feel free to let me know, and if you need to close the pull request it is also fine, in my case, I think I managed to make the assertion directly in the `<div role =" option> `, which in` <div role = "listbox"> `, I just thought it necessary to contribute, thank's so much for the review. Regards

gnapse:
> I understand, if is necessary to modify the pull request, so that suits the needs of the library, feel free to let me know, and if you need to close the pull request it is also fine,

I'm going to close it, as I do not see an alternative to solve this generically. Custom components need to adapt to the platform, and not the other way around.

Thanks anyway for your contribution. In the future, I recommend that you propose changes first by raising a discussion (an issue or a question) so you do not waste time implementing something that's potentially not suitable.

> in my case, I think I managed to make the assertion directly in the `<div role="option"></div>`, which in` <div role="listbox"></div>`, I just thought it necessary to contribute, thank's so much for the review. Regards

I'm glad that worked out for you. That's the way to go with custom stuff: custom app-specific approaches to testing as well.

PR #422, opened 2021-11-23
Fix wrong deprecate error message

MLKiiwy:
Update the deprecate error message that is thrown when using `toHaveDescription`


gnapse (approved):

gnapse:
@all-contributors, please add @MLKiiwy for tests, bug

Issue #209, opened 2020-02-18
`toBeVisible` not working as expected with `visibility` property


lhrinaldi:
- `@testing-library/jest-dom` version: v5.1.1
- `@testing-library/react` version: v9.4.0
- `node` version: v12.14.1
- `npm` (or `yarn`) version: v1.21.1

### Relevant code or config:

```js
import styled from 'styled-components';

const Container = styled.div`
	visibility: hidden;

	&:hover {
		visibility: visible;
	}

	> input:checked {
		visibility: visible;
	}
`;


const { getByTestId } = render(
	<Container>
		<input data-testid="checkbox" type="checkbox" checked />
	</Container>
);

expect(getByTestId('checkbox')).toBeVisible();
```

### What you did:

Tried to test a checkbox that must be visible when it has been checked with `toBeVisible`.

### What happened:

Even though the checkbox is visible `toBeVisible` throws an error because it detects that the parent has `visibility: hidden`.

### Reproduction:

https://codesandbox.io/s/kind-rain-62otm?fontsize=14&hidenavigation=1&previewwindow=tests&theme=dark

### Problem description:

My code has a checkbox that must be visible when the user checks it. The visibility property is perfect for this because it overrides its parents' values. 

So, if the checkbox has `visibility: visible` property, it will be shown regardless of the parent value.

### Suggested solution:

Not sure :( will come up with something if this is really an issue.

eps1lon:
The codesandbox does not use any of the code you described. Don't forget to save and freeze the codesandbox after submission.

I suspect this is an issue with `jsdom` not implementing CSS cascade. I need a repro to verify though.

r-inaldi:
@eps1lon thanks for letting me know, just included a test example there.

eps1lon:
Perfect. This is indeed a bug. 

We're currently assuming that `visibility` acts like `display`. While `display: none` hides any child even if they do set `display: block`, `visibility: hidden` can be overridden. The difference between `display` and `visibility` is that `visibility` is an inherited CSS property while `display` is not. 

We need to account for that now that JSDOM implements CSS inheritance.

r-inaldi:
I've tried to implement but I think JSDOM CSS inheritance doesn't work correctly. By default all elements have `visibility: visible`, so:

```html
<section style="display: block; visibility: hidden">
    <p>Hello <strong>World</strong></p>
</section>
```

[...]

PR #428, opened 2022-01-13
Fix is visible visibility

gnapse:
**What**:

Fixes https://github.com/testing-library/jest-dom/issues/209#issuecomment-1012250013

**Why**:

Because  `.toBeVisible(element)` is not consistent with how CSS `visibility` works.

**How**:

By fixing the logic that calculates how visibility is determined. Enough details are present in the new code in the form of comments explaining the logic.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/to-be-visible.js
+++ b/src/__tests__/to-be-visible.js
@@ -2,44 +2,145 @@ import {render} from './helpers/test-utils'
 import document from './helpers/document'
 
 describe('.toBeVisible', () => {
-  it('returns the visibility of an element', () => {
+  it('considers elements to be visible by default', () => {
     const {container} = render(`
       <div>
         <header>
-          <h1 style="display: none">Main title</h1>
-          <h2 style="visibility: hidden">Secondary title</h2>
-          <h3 style="visibility: collapse">Secondary title</h3>
-          <h4 style="opacity: 0">Secondary title</h4>
-          <h5 style="opacity: 0.1">Secondary title</h5>
+          <h1>This is the title</h1>
         </header>
-        <button hidden>Hidden button</button>
-        <section style="display: block; visibility: hidden">
+        <button>Hidden button</button>
+        <section>
           <p>Hello <strong>World</strong></p>
         </section>
       </div>
     `)
 
     expect(container.querySelector('header')).toBeVisible()
-    expect(container.querySelector('h1')).not.toBeVisible()
-    expect(container.querySelector('h2')).not.toBeVisible()
-    expect(container.querySelector('h3')).not.toBeVisible()
-    expect(container.querySelector('h4')).not.toBeVisible()
-    expect(container.querySelector('h5')).toBeVisible()
-    expect(container.querySelector('button')).not.toBeVisible()
-    expect(container.querySelector('strong')).not.toBeVisible()
+    expect(container.querySelector('h1')).toBeVisible()
+    expect(container.querySelector('button')).toBeVisible()
+    expect(container.querySelector('strong')).toBeVisible()
 
     expect(() =>
       expect(container.querySelector('header')).not.toBeVisible(),
     ).toThrowError()
-    expect(() =>
-      expect(container.querySelector('p')).toBeVisible(),
-    ).toThrowError()
   })
 
-  test('detached element is not visible', () => {
-    const subject = document.createElement('div')
-    expect(subject).not.toBeVisible()
-    expect(() => expect(subject).toBeVisible()).toThrowError()
+  describe('with the "hidden" attribute', () => {
+    it('considers an element to not be visible', () => {
+      const {container} = render('<button hidden>Click me</button>')
+      expect(container.querySelector('button')).not.toBeVisible()
+      expect(() =>
+        expect(container.querySelector('button')).toBeVisible(),
+      ).toThrowError()
+    })
+  })
+
+  describe('display', () => {
+    it.each([
+      ['inline'],
+      ['block'],
+      ['inline-block'],
+      ['flex'],
+      ['inline-flex'],
+      ['grid'],
+      ['inline-grid'],
+    ])('considers "display: %s" as visible', display => {
+      const {container} = render(`
+        <div style="display: ${display}">
+          <button>Click me</button>
+        </div>,
+      `)
+      expect(container.querySelector('button')).toBeVisible()
+      expect(() =>
+        expect(container.querySelector('button')).not.toBeVisible(),
+      ).toThrowError()
+    })
+
+    it('considers "display: none" as not visible', () => {
+      const {container} = render(`
+        <div style="display: none">
+          <button>Click me</button>
+        </div>,
+      `)
+      expect(container.querySelector('button')).not.toBeVisible()
+      expect(() =>
+        expect(container.querySelector('button')).toBeVisible(),
+      ).toThrowError()
+    })
+  })
+
+  describe('visibility', () => {
+    it('considers "visibility: collapse" as visible', () => {
+      const {container} = render(`
+        <div style="visibility: visible">
+          <button>Click me</button>
+        </div>,
+      `)
+      expect(container.querySelector('button')).toBeVisible()
+      expect(() =>
+        expect(container.querySelector('button')).not.toBeVisible(),
+      ).toThrowError()
+    })
+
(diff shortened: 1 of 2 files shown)


src/to-be-visible.js
```diff
@@ -1,33 +1,63 @@
...
+/**
+ * Computes the boolean value that determines if an element is considered visible from the
+ * `toBeVisible` custom matcher point of view.
+ *
+ * Visibility is controlled via two different sets of properties and styles.
+ *
+ * 1. One set of properties allow parent elements to fully controls its sub-tree visibility. This
+ *    means that if higher up in the tree some element is not visible by this criteria, it makes the
+ *    entire sub-tree not visible too, and there's nothing that child elements can do to revert it.
+ *    This includes `display: none`, `opacity: 0`, the presence of the `hidden` attribute`, and the
```
eps1lon, line 45: Note that you can have `<div hidden style="display: block" />` which will visible but may be excluded from the a11y tree. It should be documented if `display` or `hidden` has priority in your implementation.

eps1lon, line 45: I wouldn't include `opacity` to be honest. 0.00001 will suddenly be considered visible? Opacity also has no relevance to a11y tree inclusion.

src/to-be-visible.js
```diff
@@ -1,33 +1,63 @@
...
+ *    open state of a details/summary elements pair.
+ *
+ * 2. The other aspect influencing if an element is visible is the CSS `visibility` style. This one
```
eps1lon, line 48: nit:

```suggestion
 * 2. The other aspect influencing if an element is visible is the CSS `visibility` property. This one
```
That's how these are named officially i.e. in the spec.

src/to-be-visible.js
```diff
@@ -1,33 +1,63 @@
...
+ *    is also inherited. But unlike the previous case, this one can be reverted by child elements.
+ *    A parent element can set its visibiilty to `hidden` or `collapse`, but a child element setting
+ *    its own styles to `visibility: visible` can rever that, and it makes itself visible. Hence,
+ *    this criteria needs to be checked independently of the other one.
+ *
+ * Hence, the apprach taken by this function is two-fold: it first gets the first set of criteria
+ * out of the way, analyzing the target element and up its tree. If this branch yields that the
+ * element is not visible, there's nothing the element could be doing to revert that, so it returns
+ * false. Only if the first check is true, if proceeds to analyze the `visibility` CSS.
```
eps1lon, line 57: That's an interesting trade-off. You're basically assuming that checking `visibility` is more expensive than walking up the tree and checking `display` etc on every parent.

How was this trade-off motivated? 

In the other Testing Library implementations we're checking `visibility` on the element first since that allows us to potentially bail out of walking up the tree i.e.
```diff
 function isElementVisible(element) {
-  return isElementTreeVisible(element) && isElementVisibilityVisible(element)
+  return isElementVisibilityVisible(element) && isElementTreeVisible(element)
 }
```

src/to-be-visible.js
```diff
@@ -1,33 +1,63 @@
...
-    display !== 'none' &&
-    visibility !== 'hidden' &&
-    visibility !== 'collapse' &&
-    opacity !== '0' &&
-    opacity !== 0
-  )
+function isVisibleSummaryDetails(element, previousElement) {
+  return element.nodeName === 'DETAILS' &&
+    previousElement.nodeName !== 'SUMMARY'
+    ? element.hasAttribute('open')
```
eps1lon, line 13: question not directly related to this PR: Is the attribute or property relevant i.e. when I set `someDetailsElement.open = false` will `element.hasAttribute('open')` return `true`?

src/to-be-visible.js
```diff
@@ -1,33 +1,63 @@
 import {checkHtmlElement} from './utils'
 
-function isStyleVisible(element) {
+function getElementVisibilityStyle(element) {
+  if (!element) return 'visible'
   const {getComputedStyle} = element.ownerDocument.defaultView
+  const {visibility} = getComputedStyle(element)
+  return visibility || getElementVisibilityStyle(element.parentElement)
```
eps1lon, line 7: Why do you walk up the tree here? `visibility` should never be falsy unless the user agent has an incorrect implementation of the CSS spec. [`visibility` has an initial value of `visible`](https://developer.mozilla.org/en-US/docs/Web/CSS/visibility#formal_definition)

src/to-be-visible.js
```diff
@@ -1,33 +1,63 @@
 import {checkHtmlElement} from './utils'
 
-function isStyleVisible(element) {
+function getElementVisibilityStyle(element) {
+  if (!element) return 'visible'
```
eps1lon, line 4: This is a bit odd. If the `element` is falsy it should be considered "visible"?

src/to-be-visible.js
gnapse: Ok, but that's a separate discussion. Right now this is not new functionality, and I'm merely maintaining backward compatibility.

I agree that `0.0001` could be considered as such, but this was a compromise, because there's no arguing that `opacity: 0` makes things not visible. `toBeVisible` was never about the presence of things in the a11y tree. It was more about actual visibility with your eyes.

I'll open an issue to discuss this. But I will not remove in this PR, as it will be a breaking change.

src/to-be-visible.js
```diff
@@ -1,33 +1,63 @@
...
+ * Visibility is controlled via two different sets of properties and styles.
+ *
+ * 1. One set of properties allow parent elements to fully controls its sub-tree visibility. This
+ *    means that if higher up in the tree some element is not visible by this criteria, it makes the
+ *    entire sub-tree not visible too, and there's nothing that child elements can do to revert it.
+ *    This includes `display: none`, `opacity: 0`, the presence of the `hidden` attribute`, and the
+ *    open state of a details/summary elements pair.
+ *
+ * 2. The other aspect influencing if an element is visible is the CSS `visibility` style. This one
+ *    is also inherited. But unlike the previous case, this one can be reverted by child elements.
```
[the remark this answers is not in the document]
gnapse: Maybe "inherited" is not the best word. What I mean is that setting `display: none`, `opacity: 0`, or `visibility: hidden` in a parent element always hides the child elements too. But, unlike with the former two, the last one is reversible: child elements can make themselves visible over its parent's wish.

I'll reword to not use the word inherit to avoid confusion.

src/to-be-visible.js
gnapse: I was honestly not thinking of performance, and the comment talks more about the logic we follow. Both branches walk up the tree, and either of these branches could bail out more quickly than the other. There's no way to tell.

Though the check for visibility is simpler, so maybe you're right that on average it could be faster. I'm ok with making this change.

src/to-be-visible.js
gnapse: I'm honestly not sure. I did not even implement the support for summary/detail. It may be worth checking.

src/to-be-visible.js
gnapse: Hmmm, good point. This clarifies your point about bailing out quickly. I'll make the change.

src/to-be-visible.js
eps1lon: Makes sense. I missed, that you already included opacity.

src/to-be-visible.js
eps1lon: Yeah, for `visibility` you don't need to walk up since it the **computed** `visbility` of parents doesn't matter. only the **computed** `visibility` of the element asserted on matters.

This was relevant in earlier versions of JSOM where `visibility` wasn't inherited.

src/__tests__/to-be-visible.js
```diff
@@ -2,44 +2,181 @@ import {render} from './helpers/test-utils'
...
-      expect(container.querySelector('p')).toBeVisible(),
-    ).toThrowError()
   })
 
-  test('detached element is not visible', () => {
-    const subject = document.createElement('div')
-    expect(subject).not.toBeVisible()
-    expect(() => expect(subject).toBeVisible()).toThrowError()
+  describe('with the "hidden" attribute', () => {
+    it('considers an element to not be visible', () => {
```
aarowman, line 29:
```suggestion
    it('considers an element as not visible', () => {
```

PR #431, opened 2022-01-31
Fix wrong toHaveValue example

cbroeren:
**What**:

Small fix on README.md

**Why**:

The `.toHaveValue` example for the select verifies for `.not.toHaveValue`, but the code example has these two options selected. That means it _should_ have this value.

**How**:

I removed the `.not` part to have the test example match the code example.

**Checklist**:

- [x] Documentation
- [ ] ~Tests~ N/A
- [ ] ~Updated Type Definitions~ N/A
- [x] Ready to be merged
      

Screen Shot 2022-01-31 at 19 33 58

### DIFF
--- a/README.md
+++ b/README.md
@@ -890,7 +890,7 @@ const selectInput = getByTestId('select-number')
 expect(textInput).toHaveValue('text')
 expect(numberInput).toHaveValue(5)
 expect(emptyInput).not.toHaveValue()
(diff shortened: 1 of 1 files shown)


gnapse (approved):

gnapse:
@all-contributors please add @cbroeren for docs

PR #379, opened 2021-06-12
feat: Drop support for Node 12

nickmccurdy:
**What**: Don't test with node 10, 12, or 15

**Why**: Node 10, 12, and 15 reached EOL so let's free up some CI resources.

**How**: Remove 10, 12, and 15 from node test matrix.

**Checklist**:

- ~[ ]~ Documentation added to the
      [docs site](https://github.com/testing-library/testing-library-docs)
- [x] Tests
- ~[ ]~ Typescript definitions updated
- [x] Ready to be merged

PR #438, opened 2022-02-24
Publish both esm and cjs

### LINKED ISSUE #437: Publish ES modules

I am a maintainer of storybook-builder-vite, which allows storybook users to build their previews using vite, which works entirely in ESM.  Storybook recently also released [@storybook/jest](https://github.com/storybookjs/jest) and the latest release extends expect with jest-dom matchers.  However, since jest-dom is only published as commonjs, typescript needs to add a compatibility layer so that `import * as matchers from @testing-library/jest-dom/matchers;` can be used.  This is causing an issue, whereby a `default` is being included as well, causing expect to throw an error.  See https://github.com/storybookjs/jest/issues/15#issuecomment-1049045296 for the error description.   

To avoid the esm/cjs bundling problems, I'd like to request that jest-dom publish an esm version alongside cjs.  

### Suggested implementation:

I see that you're using `kcd-scripts build`, which also allows a `--bundle` flag that builds cjs, esm, umd, and umd.min bundles.  Alternatively, https://github.com/unjs/unbuild is a very simple build tool which can generate multiple targets.

I'm trying to troubleshoot the symptom of the esm->cjs namespace import to see if there's a way I can work around it in the storybook packages, but I think it would be nice if this package could publish esm anyway, especially now that jest itself is publishing es modules.

I think as long as the `package.json` is updated correctly, there shouldn't be any user-facing changes needed.  That said, bundling the files as suggested above will cause problems if users are importing files directly from `node_modules/@testing-library/jest-dom/dist`.  If that's the case, and it's a use case that needs to be supported, then a different build process will be needed, but can still be accomplished.

IanVS:
**What**:

This publishes both es modules and commonjs in the bundle.  

**Why**:

Closes #437 

**How**:

I used rollup to generate the two sets of bundles.  I'm not doing any processing on either of them aside from what rollup does by default, mostly because I'm not sure what level of browser/node support this project follows.  It would be pretty straightforward to add https://www.npmjs.com/package/@rollup/plugin-babel or otherwise configure the output as needed.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged
      

I haven't updated any documentation or tests, because I wanted to see if this was at all on the right track.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,14 +2,25 @@
   "name": "@testing-library/jest-dom",
   "version": "0.0.0-semantically-released",
   "description": "Custom jest matchers to test the state of the DOM",
-  "main": "dist/index.js",
+  "main": "dist/cjs/index.js",
+  "module": "dist/esm/index.js",
+  "exports": {
+    ".": {
+      "require": "./dist/cjs/index.js",
+      "import": "./dist/esm/index.js"
+    },
+    "./matchers": {
+      "require": "./dist/cjs/matchers.js",
+      "import": "./dist/esm/matchers.js"
+    }
+  },
   "engines": {
     "node": ">=8",
     "npm": ">=6",
     "yarn": ">=1"
   },
   "scripts": {
-    "build": "kcd-scripts build",
+    "build": "rollup -c",
     "format": "kcd-scripts format",
     "lint": "kcd-scripts lint",
     "setup": "npm install && npm run validate -s",
@@ -46,7 +57,8 @@
     "jest-watch-select-projects": "^2.0.0",
     "jsdom": "^16.2.1",
     "kcd-scripts": "^11.1.0",
-    "pretty-format": "^25.1.0"
+    "pretty-format": "^25.1.0",
+    "rollup": "^2.68.0"
   },
   "eslintConfig": {
     "extends": "./node_modules/kcd-scripts/eslint.js",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,0 +1,23 @@
+import path from 'path'
+import pkg from './package.json'
+
+export default [
+  {
+    input: {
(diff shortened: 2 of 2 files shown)


IanVS:
To be safe, maybe this could be released along with https://github.com/testing-library/jest-dom/pull/379 in a major version bump.

IanVS:
@gnapse I have tested in my own vite storybook app, by running `npm build`, and copying the `package.json` and `dist` over to my project's node_modules.  

I don't have a traditional jest project running in node to test the cjs build on, but the process would be the same.  

Perhaps you could publish an `alpha` or `beta` version, that people could test out?

IanVS:
@gnapse do you still plan to release a beta version with this change?  Is there anything else you'd like me to do before you're comfortable with it?

gnapse:
@IanVS sorry for the delay. I'm going to resume work on this tomorrow. I'll decide then if I go with a alpha or beta version first.

yannbf:
Thank you all for doing this! I was wondering whether there are plans to release this in an alpha version somewhere?

Thank you <3

IanVS:
Hi is there any plan to release this?  It's kind of a bummer that I put the work into it and there's still no way to use it.

yannbf:
Hey @gnapse was this change ever included in a release?

kasperpeulen:
Can this PR also be merged to main?

IanVS:
It looks like the other changes that were in `next` were cherry-picked back to main and released, but not this one.  Does anyone have any information why?  It would be great to publish ESM, if at all possible.

IanVS:
OK, I see now this needs to be reworked a bit based on the multiple export paths now supported for vitest, etc.  I'll open a new PR.

PR #443, opened 2022-03-04
Internalizes parser function from css dependencie

mechamobau:
**What**:

Internalizes CSS parser function from `css` dependencie.

**Why**:

Because the library has been [not maintained anymore](https://github.com/reworkcss/css/pull/157). And this lib uses another dependency called `source-maps-resolver` (that is not consumed by `@testing-library/jest-dom`) that is [officially deprecated](https://github.com/lydell/source-map-resolve#deprecated) too, showing an annoying warning when `jest-dom` is used.

image

**How**:

Extracting the `parse` function from [`css`](https://github.com/reworkcss/css/blob/master/lib/parse/index.js)

### DIFF
--- a/src/css-parse.js
+++ b/src/css-parse.js
@@ -1,0 +1,603 @@
+// http://www.w3.org/TR/CSS21/grammar.html
+// https://github.com/visionmedia/css-parse/pull/49#issuecomment-30088027
+var commentre = /\/\*[^*]*\*+([^/*][^*]*\*+)*\//g
+
+module.exports = function(css, options){
+  options = options || {};
+
+  /**
+   * Positional.
+   */
+
+  var lineno = 1;
+  var column = 1;
+
+  /**
+   * Update lineno and column based on `str`.
+   */
+
+  function updatePosition(str) {
+    var lines = str.match(/\n/g);
+    if (lines) lineno += lines.length;
+    var i = str.lastIndexOf('\n');
+    column = ~i ? str.length - i : column + str.length;
+  }
+
+  /**
+   * Mark position and patch `node.position`.
+   */
+
+  function position() {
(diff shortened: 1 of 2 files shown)


gnapse:
Hmmm, I'm hesitant to just bring in that entire library's core into our code as a solution. A few reasons come to mind:

- First thing I'd do is see if there are other libraries that could help us without us having to bring in this code to be maintained by ourselves. Accepting the burden to maintain this is too much.
- If we really need to fork/copy over that library's code into ours, some more considerations come to mind:
    - Lack of tests (which could be solved by bringing in the tests from that lib too)
    - Too brad scope: that code has support for parsing the entire CSS language. We only really need to parse the rules inside a selector. I'd rather trim their code down to the parts that take care of this, or get inspired by how they do it and do that ourselves. And then use/adapt their very tests on this to bring them to our library.

I know these are all more difficult in the long run than merely copying this over blindly right now. But they'll pay off in the long run.

MichaelDeBoey:
@gnapse What do you think would be the best approach on pushing this forward?

Asphiii:
Any news on that?

suutari:
FYI there seems to be a new fork of the `css` lib in progress. See https://github.com/reworkcss/css/issues/164
and https://github.com/adobe/css-tools. Maybe this `@adobe/css-tools` would be better maintained? Though currently it seems to be in Release Candidate phase still.

holblin:
The version `4.0.0` has been released today :-)
https://www.npmjs.com/package/@adobe/css-tools

This doesn't contain the source-map support but otherwise, it is a 1o1 replacement with up to 25% performance improvement.

MichaelDeBoey:
Duplicate of #470

PR #453, opened 2022-04-28
Use .mjs extension for esm files

IanVS:
**What**:

This uses `.mjs` for esm files being exported from the package, since the package itself is not `"type": "module"`.  I also marked it as explicitly `"type": "commonjs"` for good measure.

Also, I added a step to clean out the `dist` directory before each build.

And combined the `dist/esm` and `dist/cjs` directories, since now the files use different extensions.

**Why**:

https://github.com/testing-library/jest-dom/issues/437#issuecomment-1086939455

**How**:

Adjusted the rollup config

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged
      

I've opened this against the `next` branch, so that an alpha version can be released.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,19 +1,30 @@
-import path from 'path'
 import pkg from './package.json'
 
 export default [
   {
-    input: {
-      index: 'src/index.js',
-      matchers: 'src/matchers.js',
-    },
+    input: 'src/index.js',
     output: [
       {
-        dir: path.dirname(pkg.exports['./matchers'].import),
+        file: pkg.exports['.'].import,
         format: 'esm',
       },
       {
-        dir: path.dirname(pkg.exports['./matchers'].require),
+        file: pkg.exports['.'].require,
+        format: 'cjs',
+      },
+    ],
+    external: id =>
+      !id.startsWith('\0') && !id.startsWith('.') && !id.startsWith('/'),
+  },
+  {
+    input: 'src/matchers.js',
+    output: [
+      {
+        file: pkg.exports['./matchers'].import,
+        format: 'esm',
+      },
+      {
+        file: pkg.exports['./matchers'].require,
         format: 'cjs',
       },
     ],
--- a/package.json
+++ b/package.json
@@ -2,16 +2,17 @@
   "name": "@testing-library/jest-dom",
   "version": "0.0.0-semantically-released",
   "description": "Custom jest matchers to test the state of the DOM",
-  "main": "dist/cjs/index.js",
-  "module": "dist/esm/index.js",
+  "main": "dist/index.js",
+  "module": "dist/index.mjs",
+  "type": "commonjs",
   "exports": {
     ".": {
-      "require": "./dist/cjs/index.js",
-      "import": "./dist/esm/index.js"
+      "require": "./dist/index.js",
+      "import": "./dist/index.mjs"
     },
     "./matchers": {
-      "require": "./dist/cjs/matchers.js",
-      "import": "./dist/esm/matchers.js"
+      "require": "./dist/matchers.js",
+      "import": "./dist/matchers.mjs"
     }
   },
   "engines": {
@@ -20,7 +21,7 @@
     "yarn": ">=1"
   },
   "scripts": {
-    "build": "rollup -c",
+    "build": "rimraf dist && rollup -c",
     "format": "kcd-scripts format",
     "lint": "kcd-scripts lint",
     "setup": "npm install && npm run validate -s",
@@ -58,6 +59,7 @@
     "jsdom": "^16.2.1",
     "kcd-scripts": "^11.1.0",
     "pretty-format": "^25.1.0",
+    "rimraf": "^3.0.2",
     "rollup": "^2.68.0"
   },
   "eslintConfig": {


IanVS:
Hi @gnapse, @sheremet-va, how's this look?

IanVS:
Any feedback on this?

rollup.config.js
```diff
@@ -1,19 +1,30 @@
-import path from 'path'
 import pkg from './package.json'
 
 export default [
   {
-    input: {
-      index: 'src/index.js',
-      matchers: 'src/matchers.js',
-    },
+    input: 'src/index.js',
```
sheremet-va, line 5: Maybe it's better to provide an array for an input? So rollup can share code between files, otherwise matchers are duplicated.

sheremet-va (changes requested):

rollup.config.js
IanVS: I'm not terribly familiar with rollup.  How can I specify an array for input, when each one should have a specific different output?

sheremet-va: There is an example in Vitest repo: https://github.com/vitest-dev/vitest/blob/main/packages/vitest/rollup.config.js

You will need to entire, something like this:

```ts
export default [
{
  input: entries,
  output: {
    dir: 'dist',
    entryFileNames: '[name].mjs',
   format: 'esm'
  }
},
{
  input: entries,
  output: {
    dir: 'dist',
    entryFileNames: '[name].js',
   format: 'cjs'
  }
}
]
```

rollup.config.js
```diff
@@ -1,22 +1,23 @@
...
 
 export default [
   {
-    input: {
-      index: 'src/index.js',
-      matchers: 'src/matchers.js',
+    input: entries,
+    output: {
+      dir: 'dist',
+      entryFileNames: '[name].mjs',
```
sheremet-va, line 8: You can add `chunkFileNames: '[name]-[hash].mjs'`, so chunks will also have correct filepath.

You can also add `preserveModules: true`, then rollup will not generate "chunks", it will have the same file structure as your source code. 

I guess the first solution is closest to how it was, so 😄

sheremet-va (changes requested):
Rollup now produces `.js` file with ESM content alongside correct `.mjs` file 😞 

This file _should_ be generated, so you don't have double code for `/index` and `/matchers` entry points.

I've added two possible solutions.

rollup.config.js
IanVS: Good eye, thanks for catching that.  I've taken your first suggestion, and mirrored it on the .js as well, just for consistency / to be explicit.

sheremet-va (changes requested):
I've tried this file, and it unfortunately fails with folder error :(

```
// test.mjs
import {toBeChecked} from './dist/matchers.mjs'

console.log(toBeChecked)
```

IanVS:
It works for me if I add the `--experimental-specifier-resolution=node` flag.  

```
node --experimental-specifier-resolution=node ./test.mjs
```

I don't work directly in node.js much, and espeically not with esm, but I do know I've needed to add this flag in the past.  Is that not what you normally do?

sheremet-va:
> It works for me if I add the `--experimental-specifier-resolution=node` flag.
> 
> ```
> node --experimental-specifier-resolution=node ./test.mjs
> ```
> 
> I don't work directly in node.js much, and espeically not with esm, but I do know I've needed to add this flag in the past. Is that not what you normally do?

Well, it will be removed when Loader APIs will be finalized. I _guess_ the problem is the other package that doesn't define `exports` (so, ecosystem problem), or `jest-dom` can import directly from `index.js` file.

IanVS:
Boy, ESM in node is kind of a PITA isn't it?  lol.

I added extensions / index.js, and also handled a file that I guess I had missed previously, extend-expect.js.  I think using exports like this will work, but I'm not sure if I need the extension in the export map, or not.

sheremet-va:
> Boy, ESM in node is kind of a PITA isn't it? lol.

Yes, ESM<->CJS is a crime against humanity 😢  

> I think using exports like this will work, but I'm not sure if I need the extension in the export map, or not.

If you want people to import `@testing-library/dom/extend-expect`, it will not work 👀

sheremet-va:
I see that everything else is working fine. But I thought you should know:
```ts
import '@testing-library/jest-dom/extend-expect' // doesn't work
import '@testing-library/jest-dom/extend-expect.js' // works
const def = require('@testing-library/jest-dom/extend-expect') // doesn't work
```

sheremet-va:
I think it might be a good idea to add a test for simply importing `jest-dom` 😄

IanVS:
@gnapse @nickmccurdy what is the purpose of `extend-expect`?  If we're talking about making breaking changes for the next major anyway, is there a reason to keep this extra export around, that seems to have just re-exported the main index file anyway?

nickmccurdy:
I assume it's just there for back compatibility. I agree that we probably don't need to keep it around for the next major (as long as we move the one-line implementation back to `index.js`).

IanVS:
@nickmccurdy, done, thanks.  I believe this is ready for review.

sheremet-va (approved):
LGTM 👍🏻

IanVS:
@nickmccurdy @gnapse  Hi, what are your thoughts on merging this and cutting a beta soon?

IanVS:
Hi, it's a bit of a shame that a breaking change was released without this PR and the related change to build and publish ESM.  Was there a reason not to include them?  Is another breaking change on the horizon?  @nickmccurdy @gnapse @kentcdodds. 🙏

kentcdodds:
Sorry, there's a skeleton crew maintaining this project. That's why this wasn't merged. I'm afraid I only have the bandwidth to work on things that directly impact my own work.

kentcdodds:
I am actually, but with vitest and I don't think I've had issues. I'm going to look at it again in the future and if this fixes things for me then I may merge your other PR unless another maintainer steps up before I get to it

jgoz:
@IanVS I think dual-publishing ESM/CJS could (and should?) be done in a non-breaking way, especially now that we've already done one breaking change that is creating some confusion amongst users (😅).

nickmccurdy:
You're right, I got mixed up when reviewing other notifications here.

PR #470, opened 2022-07-15
Migrate ccs v3 to @adobe/css-tools v4

holblin:
**What**:

Migrate the dependency `css` to `@adobe/css` to reduce maintained dependencies.

**Why**:

CSS package is not maintained anymore and has some sub-dependencies which are deprecated.

**How**:

Change the import

**Checklist**:

- [X] Documentation N/A
- [X] Tests N/A
- [X] Updated Type Definitions N/A
- [X] Ready to be merged

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,6 +1,6 @@
 import redent from 'redent'
-import {parse} from '@adobe/css-tools/dist/cjs/cssTools'
 import isEqual from 'lodash/isEqual'
+import {parse} from '@adobe/css-tools'
 
 class GenericTypeError extends Error {
   constructor(expectedString, received, matcherFn, context) {
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,5 +1,5 @@
 import redent from 'redent'
-import cssParse from 'css/lib/parse'
+import {parse} from '@adobe/css-tools/dist/cjs/cssTools'
 import isEqual from 'lodash/isEqual'
 
 class GenericTypeError extends Error {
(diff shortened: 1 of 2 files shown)


holblin:
Note: The version 4 of `@adobe/css-tools` is very recent. I used `import {parse} from '@adobe/css-tools/dist/cjs/cssTools'` to make sure it load the `cjs` build and not the `umd` build. If someone knows why `jest-dom` try to load the `umd` build instead of the `cjs` and as a solution, feel free to share :-)

MichaelDeBoey (approved):

kentcdodds (approved):
LGTM, but I'm not a current maintainer. If nobody merges in the next week or so though I'll go ahead and merge

src/utils.js
```diff
@@ -1,5 +1,5 @@
 import redent from 'redent'
-import cssParse from 'css/lib/parse'
```
xfix, line 2: Directly loading `dist/cjs/cssTools` feels somewhat hacky as there is no guarantee this path will exist in future releases of `@adobe/css-tools`. I wonder if using `require` instead would work.

holblin: I agree but I can't found a way to load the CJS build.

If I do without a path:
```
    ReferenceError: self is not defined

      1 | import redent from 'redent'
    > 2 | import {parse} from '@adobe/css-tools'
        | ^
```
If I do a require:
```
   ReferenceError: self is not defined

      1 | import redent from 'redent'
      2 | import isEqual from 'lodash/isEqual'
    > 3 | const {parse} = require('@adobe/css-tools')
        |                 ^
```

Let me know if you have a better solution or something that I could change for the build of `@adobe/css-tools`

holblin: I made a change of how `@adobe/css-parse` is build (UMD build, globalThis is now set to this in wepack config). This should solve the issue/concern.

holblin:
@kentcdodds I updated the import to solve the CJS import path that I was using :-)
Could you merge it when you get some time? 😇 Thanks!

kentcdodds (approved):
Thank you 👍

PR #472, opened 2022-08-29
chore: remove `@types` from dependencies

### LINKED ISSUE #427: TS error when using vitest globals and @testing-library/jest-dom
- `@testing-library/jest-dom` version: 5.16.1
- `node` version: 14.17.0
- `yarn` version: 1.22.17
- `@testing-library/react` version: 12.1.2

I am using `vitest` instead of `jest` for testing a React component lib in a monorepo. I am using `@testing-library/jest-dom` and `@testing-library/react` and both happily work with `vitest`.

However, when type checking my code I am seeing a clash between `@types/jest` and `vitest`. The `@types/jest` dependency seems to have been pulled in by  `@testing-library/jest-dom`.

```
../node_modules/@types/jest/index.d.ts:34:1 - error TS6200: Definitions of the following identifiers conflict with those in another file: test, describe, it, expect, beforeAll, afterAll, beforeEach, afterEach

34 declare var beforeAll: jest.Lifecycle;
   ~~~~~~~

  ../node_modules/vitest/global.d.ts:1:1
    1 declare global {
      ~~~~~~~
    Conflicts are in this file.

../node_modules/vitest/global.d.ts:1:1 - error TS6200: Definitions of the following identifiers conflict with those in another file: test, describe, it, expect, beforeAll, afterAll, beforeEach, afterEach

1 declare global {
  ~~~~~~~

  ../node_modules/@types/jest/index.d.ts:34:1
    34 declare var beforeAll: jest.Lifecycle;
       ~~~~~~~
    Conflicts are in this file.


Found 2 errors.
```

It would be great to have a workaround for this issue so the `jest` types can be removed or ignored.

To reproduce this issue clone the following project (notice it uses a branch called `yarn`):

https://github.com/robcaldecott/pnpm-vite-monorepo-example/tree/yarn

To see the error:

```
yarn install
cd components
yarn build
```

wallzero:
Fix #427

**What**:

Type definitions should be under `devDepedencies`

**Why**:

Types are not runtime dependencies. It is also easier for downstream projects to manage types.

**How**:

Moved `@types/testing-library__jest-dom` from `dependencies` to `devDepedencies` in `package.json`.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,6 @@
     "redent": "^3.0.0"
   },
   "devDependencies": {
-    "@types/testing-library__jest-dom": "^5.9.1",
     "jest-environment-jsdom-sixteen": "^1.0.3",
     "jest-watch-select-projects": "^2.0.0",
     "jsdom": "^16.2.1",
--- a/package.json
+++ b/package.json
@@ -32,7 +32,6 @@
   "license": "MIT",
   "dependencies": {
     "@babel/runtime": "^7.9.2",
-    "@types/testing-library__jest-dom": "^5.9.1",
     "aria-query": "^5.0.0",
     "chalk": "^3.0.0",
     "@adobe/css-tools": "^4.0.1",
@@ -42,6 +41,7 @@
     "redent": "^3.0.0"
   },
   "devDependencies": {
+    "@types/testing-library__jest-dom": "^5.9.1",
     "jest-environment-jsdom-sixteen": "^1.0.3",
     "jest-watch-select-projects": "^2.0.0",
     "jsdom": "^16.2.1",


nickmccurdy:
We intentionally bundle the types so they don't have to be installed separately with npm. I'd also argue having these types in devDependencies is incorrect, since the types are used when Jest DOM is installed, not when it's in development.

nickmccurdy:
I've changed my mind. I think we should just uninstall the `@types` package completely to fix Vitest support (#427). We're not using it in development anyway.

Note that I'm intentionally suggesting this as a breaking change, as it accommodates using TypeScript with Vitest without conveniently. Jest users would need to install this package, and we should document that before publishing a major release.

nickmccurdy (approved):
I've changed my mind. I think we should just uninstall the `@types` package completely to fix Vitest and TypeScript support (#427). We're not using it in development anyway.

Note that I'm intentionally suggesting this as a breaking change, as it's important for Vitest and TypeScript users. Jest users would need to install this package, and we should document that before publishing a major release. I'd like another maintainer to approve this before merging ideally.

EzraBrooks:
Any idea when this change will be merged? This would be very helpful for me, as I'd like to drop `@types/jest` in favor of `@jest/globals` in my project but cannot because of this transitive dependency that extends the global namespace.

kentcdodds:
Vitest (is fantastic and) has made this challenging. I wonder whether there's an established pattern for supporting both since the matchers API in vitest is forked from Jest. I definitely feel like this needs to be solved.

Unfortunately I don't have bandwidth to work on this myself so I'm not going to merge this and put that burden on someone else, but I wanted to come in here and voice my support for doing something about this problem.

kentcdodds:
Oh, and personally, I'm actually ok with this change (of course, it would be breaking). It's more in line with the way most packages work with definitely typed anyway 🤷‍♂️

But it would be nice if there was a way to make the types themselves support both vitest and jest at the same time. I have no idea whether that's possible.

jgoz:
@kentcdodds 

> But it would be nice if there was a way to make the types themselves support both vitest and jest at the same time. I have no idea whether that's possible.

#511

Issue #457, opened 2022-05-17
Upgrade to v28 complains "Cannot find module '@jest/expect'"


tettoffensive:

heinrich321:
Having same issue in non-react app

sauhaardac:
Same here in a react app

areindl:
I had the same issue.

Do you have a jest.config.js? 

This is was fixed it for me.

I removed `resolver: 'jest-node-exports-resolver'` and changed it from:

```js
module.exports = {
    preset: 'ts-jest', // use js-jest if you re not using typescript
    testEnvironment: 'node',
    resolver: 'jest-node-exports-resolver',
}

```

to:

```js

module.exports = {
    preset: 'ts-jest',
    testEnvironment: 'node',
    transformIgnorePatterns: ['^.+\\.js$'],
}

```


Can you try?

heinrich321:
@areindl I did the same, fixed the issue +1

hunterdes:
You might not need this 'jest-node-exports-resolver' as https://github.com/facebook/jest/releases/tag/v28.0.0 already support export in package.json

Issue #426, opened 2022-01-10
Missing type definitions for expect imported from @jest/globals


arty-name:
- `@testing-library/jest-dom` version: 5.16.1
- `node` version: v16.13.1
- `yarn` version: 1.22.17

### Relevant code or config:

```js
import { expect } from '@jest/globals';
import '@testing-library/jest-dom';
expect(container).toHaveAttribute('hidden');
```

### What you did:

I imported jest globals from '@jest/globals'

### What happened:

I got the TypeScript error message: `TS2339: Property 'toHaveAttribute' does not exist on type 'Matchers '.`

### Reproduction:

```js
import { expect } from '@jest/globals';
import '@testing-library/jest-dom';
expect(container).toHaveAttribute('hidden');
```

### Problem description:

The custom matchers provided by jest-dom cannot be used in TypeScript if the `expect` is imported from `@jest/globals`. Only the global `expect` is modified, apparently.

### Suggested solution:

Extend the interface of `expect` in the `@jest/globals` as well.

gnapse:
Thanks for reporting this.

I have to say, I see this more as a problem with jest than with jest-dom, if really the solution involves duplicating the code that extends the interface of `expect`'s return type. I hope there would a better solution where you extend it once, and no matter where you get `expect` from, it should work.

arty-name:
After a deeper look, I see that the [types of jest-dom extend jest.Matchers](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/testing-library__jest-dom/index.d.ts)

Meanwhile, the [types of `@jest/globals` simply re-export the type of the `expect` package](https://github.com/facebook/jest/blob/main/packages/jest-globals/src/index.ts#L14)

I’d naively expect that it’s rather `expect.Matchers` that should be extended, but after some poking around, I feel that I am out of my depth here.

IanVS:
I'm importing `expect` directly from the `expect` package, and also struggling to get types working, since the way that jest-dom extends the types is by modifying the `jest` module, it seems.  But I don't have that, since I'm using expect outside of jest.  :(

muratx10:
Have the same issue with type definitions.

muratx10:
In my case, I've just switched to `expect` that comes from `jest`, not from `jest/globals` and it works like a charm.

rtruong:
Is there a solution for this yet?

aalpgiray:
I assumed this should worked but couldn't get it to work yet. 

[...]

Issue #314, opened 2020-12-01
Move Types from @types/* to this repo


MatanBobi:
### Describe the feature you'd like:

Following [this](https://github.com/testing-library/dom-testing-library/issues/494#issuecomment-602720977) comment and [this](https://github.com/testing-library/react-testing-library/issues/620) issue, we would like to move the types to this repo also.

### Suggested implementation:

Here's a PR done in dom-testing-library:
https://github.com/testing-library/dom-testing-library/pull/530

gnapse:
Usually there have been more difficulty with our repo than with other libs in the testing-library org because unlike other libs, this one is not used via direct imports. That usually has been a blocker.

For instance, we used to have the types baked in, and we had to extract them because it did not work properly with VSCode intellisense. See #123 for instance. There was even an effort to port the actual source code to TypeScript (see #246) and it too introduced some issues that made it not suitable for a release.

That being said, I'd be glad if we can pull it off provided it is done so that all the previous issues that we've had with this.

MatanBobi:
> #123

Thanks @gnapse, I wasn't aware of the previous attempts that led to us moving this to Definitely-Typed.
I'll give it a shot and update here :)

gnapse:
I gotta say, the barrier of entry to contribute types via having to do it in `DefinitelyTyped/DefinitelyTyped` is huge. I love TS, and I can barely drag myself to do it (cloning that huge repo or following the instructions to clone sparsely is discouraging). I really wish we could do this somehow. I will take a look at how other jest matchers libraries do it.

gnapse:
cc @testing-library/all-maintainers in case someone has some ideas on how to pull this off.

eps1lon:
If I remember correctly there wasn't much to it.

1. Add types to the original repository (SemVer minor)
2. Remove types from DefinitelyTyped

gnapse:
We may attempt it again, but the main reason we moved the type definitions from being inside the repo was #123.

There were also issues with actually writing the code in TS. See #246 (which funnily enough is 123 * 2 😄)

jdanil:
I think another reason in favour of this proposal would be compatibility with stricter node linkers. I'm finding that when using `@testing-library/jest-dom` with pnp or pnpm node linkers, the type definitions declared in this package don't actually work because they are not a dependency of the consuming project.

For example, the node modules structure may look something like this...

```
my-project
  node_modules
    @testing-library
      jest-dom
        node_modules
          @types
[...]

PR #483, opened 2022-10-06
Use expect 28 style typescript module augmentation

IanVS:
**What**:

This adds a `types` export for the package, using jest 28's recommended way to augment `expect`: https://jestjs.io/blog/2022/04/25/jest-28#expect

This might be a breaking change for users on jest 27 or below and using typescript.

**Why**:

I want to update `@storybook/expect` to use jest 28, but currently, due to the typescript changes in jest 28, the types from jest-dom are broken.  

I took a look at a few issues in this repo, including https://github.com/testing-library/jest-dom/issues/426, https://github.com/testing-library/jest-dom/issues/314, and https://github.com/testing-library/jest-dom/issues/457, and I think I'm not alone in having issues with jest 28 and types.  

**How**:

This augments the `expect` package, using the types from definitely-typed (instead of trying to move them into this package for now).  

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [X] Updated Type Definitions
- [ ] Ready to be merged
      

I'm opening this up for discussion.  I tested this approach out in my own project, and it works correctly when I also add an `import '@testing-library/jest-dom';` statement in `@storybook/expect`.  I wanted to see if the maintainers of this repo would consider making a breaking change to support jest 28+, or if you have any other ideas of how to handle types for projects which aren't using `jest` globals, but are using `expect` directly. 

Note: I also had to tweak the eslint config a little bit, so that the `@typescript-eslint` rules stopped failing and would let the precommit hook succeed.  Buuuut, it looks like that's failing CI.  Maybe a maintainer can get `main` working correctly with eslint, and I can rebase this on top?

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,9 @@
+// The actual type definitions live in definitely typed:
+import {TestingLibraryMatchers} from '@testing-library/jest-dom/matchers'
+
+declare module 'expect' {
+  interface AsymmetricMatchers
+    extends TestingLibraryMatchers<typeof expect.stringContaining, void> {}
+  interface Matchers<R>
+    extends TestingLibraryMatchers<typeof expect.stringContaining, R> {}
+}
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "0.0.0-semantically-released",
   "description": "Custom jest matchers to test the state of the DOM",
   "main": "dist/index.js",
+  "types": "index.d.js",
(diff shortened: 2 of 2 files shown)


IanVS:
@nickmccurdy @gnapse I'd love to hear what y'all think of this, or if you have any other ideas of how to handle types for those of us not using jest globally.  (e.g. https://github.com/testing-library/jest-dom/issues/427).

PR #490, opened 2022-11-24
Fix to-be-disabled reaction on `disabled="false"`

JakubGIT:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged


pseudogram:
I was going to give this a +1 then I read the [docs for the HTML attribute 'disabled'](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled).

A note for those to come:

> If you are writing semantic HTML you should not add the attribute disabled if a button is enabled. There is no such thing as `disabled=false`.

Just reading the double negative should indicate to you that something is wrong. Ite, have a good day.

PR #370, opened 2021-05-30
Add `toHaveErrorMessage` matcher

SevenOutman:
**What**:

feature: `.toHaveErrorMessage` matcher to check whether an invalid element has requested [error message](https://www.w3.org/TR/wai-aria/#aria-errormessage).

**Why**:

This PR is basically a replacement of the stale PR #258 which implements proposal #256. 

**How**:

Most implementation and tests are copied from the original PR with some modification to address testing issues. 

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Updated Type Definitions https://github.com/DefinitelyTyped/DefinitelyTyped/pull/53505
- [x] Ready to be merged

PR #503, opened 2023-07-17
Conform `toHaveErrorMessage` to Spec and Rename to `toHaveAccessibleErrorMessage`

### LINKED ISSUE #256: toHaveErrorMessage

ITenthusiasm:
## What

Primarily, this PR is intended to be a fix that conforms the original `toHaveErrorMessage` method more closely to the [WAI ARIA spec](https://www.w3.org/TR/wai-aria-1.2/#aria-errormessage). This PR also renames `toHaveErrorMessage` to `toHaveAccessibleErrorMessage`.

There was also a slight change to how the assertions are exported from `matchers.js`.

### Spec Concerns:

Currently, there are some ways in which `toHaveErrorMessage` doesn't conform to the specifications:
- `aria-errormessage` is expected to be a single `id`, not a list of `id`s. Supplying multiple `id`s is illegal (because it would constitute an invalid singular `id`). I didn't recognize this until reading through the spec more closely.
- The cases in which an element technically "does not have an accessible error message" could be improved for the sake of developer experience and expectations.

### Rename:

The other a11y-related methods explicitly call out the fact that they're intended for accessibility-related assertions. I figured it would be helpful to keep things consistent.

### Exports Update:

This is completely optional. But since it was simple, I thought it was worth trying and running by the maintainers. The new approach involves less lines of code, which will hopefully be more readable and less prone to error when it's updated.

## How

These changes were implemented with a new `toHaveAccessibleErrorMessage` method that largely built on top of the previous `toHaveErrorMessage` method. I tested the `aria-errormessage` attribute with my iPhone to make sure the test cases that I wrote were consistent with reality.

Technically speaking, it's debatable whether or not the `toHaveAccessibleErrorMessage` assertion should have early error cases that guide the user to success with intermediate checks (e.g., checks on `aria-errormessage` and `aria-invalid`). But since the `aria-errormessage` attribute is a little more complicated than the others, it could be helpful to keep the checks/warnings around. I'm not tied to down to either view, but I continued with the pattern that was already established (i.e., the pattern of helping the developer correctly give an error message to an element).

## Checklist

~~I forgot the typedefs and docs. :sob: I'll look into this. Feel free to add any comments in the meantime.~~

- [x] [Documentation](https://github.com/ITenthusiasm/jest-dom/tree/pr/fix/to-have-errormessage#tohaveaccessibleerrormessage)
- [x] Tests
- [x] [Updated Type Definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66089)
- [x] Ready to be merged

Resolves #256
Improves #370

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,7 @@ clear to read and to maintain.
   - [`toContainElement`](#tocontainelement)
   - [`toContainHTML`](#tocontainhtml)
   - [`toHaveAccessibleDescription`](#tohaveaccessibledescription)
+  - [`toHaveAccessibleErrorMessage`](#tohaveaccessibleerrormessage)
   - [`toHaveAccessibleName`](#tohaveaccessiblename)
   - [`toHaveAttribute`](#tohaveattribute)
   - [`toHaveClass`](#tohaveclass)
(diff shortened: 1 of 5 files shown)


ITenthusiasm:
So I actually don't know where the type definitions are. :flushed: Some assistance would be helpful. :sweat_smile:

PR #511, opened 2023-08-12
feat!: local types, supporting jest, @jest/globals, vitest

### LINKED ISSUE #314: Move Types from @types/* to this repo

### LINKED ISSUE #427: TS error when using vitest globals and @testing-library/jest-dom

### LINKED ISSUE #476: @types/testing-library__jest-dom ^5.9.1 has jest * in deps and it brokes yarn install
Please, set fixed version of @types/jest in deps of @types/testing-library__jest-dom because latest @types/jest 29 installs latest expect "^29.0.0". and "expect" package has node engine requirement that does not support node 12.18.1 and it broke our installation

### LINKED ISSUE #484: Update @types/testing-library__jest-dom dependency
- `@testing-library/jest-dom` version: ^5.16.5
- `node` version: 16
- `npm` (or `yarn`) version: yarn 2 latest

### Relevant code or config:

An example of when this is an issue: (see below for the root cause)

```tsx
it('should have an accessible name', () => {
  const { getByRole } = render(<input type="checkbox" aria-label="label" />);
  const inputEl = getByRole('checkbox');

  expect(inputEl).toHaveAccessibleName('label'); // ❌ toHaveAccessibleName does not exist (TS and eslint errors)
});
```

### What you did:

I'm asserting that components must have accessible names/descriptions.

### What happened:

There's a TypeScript error warning that `toHaveAccessibleName` does not exist in the Jest global matchers.

### Reproduction:

I don't think it's necessary in this case. Let me know if you'd like it.

### Problem description:

The issue is with the `@types` dependency that `@testing-library/jest-dom` includes:

https://github.com/testing-library/jest-dom/blob/948d90f32cc79339bdeebea0454599db74c5d071/package.json#L35

This is an outdated version (the latest is [5.14.5](https://www.npmjs.com/package/@types/testing-library__jest-dom/v/5.14.5) from 4 months ago) which doesn't yet declare matchers such as `toHaveAccessibleName`.

### Suggested solution:

- update the dependency 🙂 
- even better (?): include types as part of `@testing-library/jest-dom`. I've seen this suggested in several issues, but haven't dug into what the blockers would be. But including types as part of this package would prevent this kind of issues from happening in the future
- (short-term fix for people running into the issue: install the latest version of `@types/testing-library__jest-dom` in your project)

### LINKED ISSUE #506: Work out of the box with explicit imports from '@jest/globals'
We are moving away from using global imports for our tests. Partly because we use Cypress and Jest in the same packages which causes type collisions, but also just because it is much easier to reason about.

Out of the box trying to use` import '@testing-library/jest-dom'` when using the explicit imports `import { expect } from '@jest/globals'`  does not work.

I have spent a while and the best solution I have cobbled together in our local packages is to paste out parts of the types and augment, but I am sure there must be a better way!

```
import type { BaseExpect } from 'expect';
import type { addSerializer, SnapshotMatchers } from 'jest-snapshot';
import type { AsymmetricMatchers, Matchers } from '@jest/expect';
import { expect as jestGlobalExpect } from '@jest/globals';
import * as matchers from '@testing-library/jest-dom/matchers.js';

// BE AWARE this file may need updating from time to time as @testing-library/jest-dom package or @jest/globals is updated.

declare type PromiseMatchers<T = unknown> = {
  /**
   * Unwraps the reason of a rejected promise so any other matcher can be chained.
   * If the promise is fulfilled the assertion fails.
   */
  rejects: JestMatchers<Promise<void>, T> &
    Inverse<JestMatchers<Promise<void>, T>>;
  /**
   * Unwraps the value of a fulfilled promise so any other matcher can be chained.
   * If the promise is rejected the assertion fails.
   */
  resolves: JestMatchers<Promise<void>, T> &
    Inverse<JestMatchers<Promise<void>, T>>;
};

type JestMatchers<R extends void | Promise<void>, T> = Matchers<R, T> &
  SnapshotMatchers<R, T>;

declare type Inverse<Matchers> = {
  /**
   * Inverse next matcher. If you know how to test something, `.not` lets you test its opposite.
   */
  not: Matchers;
};

type JestExpect = {
  <T = unknown>(
    actual: T,
  ): JestMatchers<void, T> &
    Inverse<JestMatchers<void, T>> &
    PromiseMatchers<T> &
    matchers.TestingLibraryMatchers<unknown, T> &
    Inverse<matchers.TestingLibraryMatchers<unknown, T>>;
  addSnapshotSerializer: typeof addSerializer;
} & BaseExpect &
  AsymmetricMatchers &
  Inverse<Omit<AsymmetricMatchers, 'any' | 'anything'>>;

jestGlobalExpect.extend(matchers);

const expect = jestGlobalExpect as unknown as JestExpect;

export { expect };
```

jgoz:
**What**:

Moves matcher types back into this package and adds support for `@jest/globals` and `vitest`, both in terms of automatically extending `expect` and augmenting the appropriate TypeScript interfaces.

There are three **breaking changes** here:
1. Drop support for node < 14
2. Removes the extend-expect script. Users should use the default import path or one of the new test platform-specific paths to automatically extend the appropriate "expect" instance. `extend-expect` was not documented in the Readme, so this change should have minimal impact.
3. Remove the `@types/` dependency. This might affect _JavaScript_ users who were relying on the automatic inclusion of the `@types/` package to get auto-complete in their editor for jest-dom assertions.

But in my local testing with the changes in this PR, following the Readme was sufficient to get full auto-complete support even for JS-only users. As long as they have a jest-setup file that imports the bare export and add a dependency on `@types/jest`, recent versions of VS Code will figure it out.

In addition to the above changes, `@jest/globals` and `vitest` users can now use jest-dom without caveats. Users can now use the following import paths to automatically extend "expect" for their chosen test platform:

- @testing-library/jest-dom &mdash; jest+@types/jest
- @testing-library/jest-dom/jest-globals &mdash; @jest/globals
- @testing-library/jest-dom/vitest &mdash; vitest

For example:

```ts
import '@testing-library/jest-dom/jest-globals'
```

Importing from one of the above paths will augment the appropriate matcher interface for the given test platform, assuming the import is done in a .ts file that is included in the user's tsconfig.json.

It's also (still) possible to import the matchers directly without side effects:

```ts
import * as matchers from '@testing-library/jest-dom/matchers'
```

**Why**:

There have been numerous issues filed over the last couple years as the JS testing ecosystem has evolved and users wanted to use this library in more environments. It was usually possible, thanks to broad compatibility efforts in the ecosystem at large, but using jest-dom outside of Jest was not always straightforward.

Originally, moving the types to DT was in service of a noble "it should just work" goal, but in practice, it traded convenience in simple cases for confusion in more complex cases. It also broke with convention around DT packages &mdash; why does the actual package have a dependency on its own types? Why not simply move the types into the package? Those were valid questions that had legitimate answers, but they grew stale with time.

Fixes #314 
Fixes #427
Closes #472
Fixes #506
Fixes #476
Closes #484 
Closes #496
Closes #483
Closes #460 

**How**:

- Move latest typings from `@types/testing-library__jest-dom` into this repo
- Add `jest.d.ts` for extending global interfaces when using @types/jest
- Add `jest-globals.d.ts` for extending interfaces when using @jest/config
- Add `vitest.d.ts` for extending interfaces when using vitest
- Add optional peer dependencies on jest, @types/jest, @jest/globals, and vitest
- Add `/jest-globals` and `/vitest` exports for automatically extending @jest/globals and vitest, respectively

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,10 @@ clear to read and to maintain.
 
 - [Installation](#installation)
 - [Usage](#usage)
+  - [With `@jest/globals`](#with-jestglobals)
+  - [With Vitest](#with-vitest)
   - [With TypeScript](#with-typescript)
+  - [With another Jest-compatible `expect`](#with-another-jest-compatible-expect)
 - [Custom matchers](#custom-matchers)
   - [`toBeDisabled`](#tobedisabled)
   - [`toBeEnabled`](#tobeenabled)
@@ -128,6 +131,39 @@ import '@testing-library/jest-dom'
 setupFilesAfterEnv: ['<rootDir>/jest-setup.js']
 ```
 
+### With `@jest/globals`
+
+If you are using [`@jest/globals`][jest-globals announcement] with
+[`injectGlobals: false`][inject-globals docs], you will need to use a different
+import in your tests setup file:
+
+```javascript
+// In your own jest-setup.js (or any other name)
+import '@testing-library/jest-dom/jest-globals'
+```
+
+[jest-globals announcement]:
+  https://jestjs.io/blog/2020/05/05/jest-26#a-new-way-to-consume-jest---jestglobals
+[inject-globals docs]:
+  https://jestjs.io/docs/configuration#injectglobals-boolean
+
+### With Vitest
+
+If you are using [vitest][], this module will work as-is, but you will need to
+use a different import in your tests setup file. This file should be added to
+the [`setupFiles`][vitest setupfiles] property in your vitest config:
+
+```javascript
+// In your own vitest-setup.js (or any other name)
+import '@testing-library/jest-dom/vitest'
+
+// In vitest.config.js add (if you haven't already)
+setupFiles: ['./vitest-setup.js']
+```
+
+[vitest]: https://vitest.dev/
+[vitest setupfiles]: https://vitest.dev/config/#setupfiles
+
 ### With TypeScript
 
 If you're using TypeScript, make sure your setup file is a `.ts` and not a `.js`
@@ -142,6 +178,18 @@ haven't already:
     ...
     "./jest-setup.ts"
   ],
+```
+
+### With another Jest-compatible `expect`
+
+If you are using a different test runner that is compatible with Jest's `expect`
+interface, it might be possible to use it with this library:
+
+```javascript
+import * as matchers from '@testing-library/jest-dom/matchers'
+import {expect} from 'my-test-runner/expect'
+
+expect.extend(matchers)
 ```
 
 ## Custom matchers
--- a/jest-globals.js
+++ b/jest-globals.js
@@ -1,0 +1,4 @@
+const globals = require('@jest/globals')
+const extensions = require('./dist/matchers')
+
+globals.expect.extend(extensions)
--- a/vitest.d.ts
+++ b/vitest.d.ts
@@ -1,0 +1,1 @@
+/// <reference path="types/vitest.d.ts" />
--- a/vitest.js
+++ b/vitest.js
@@ -1,0 +1,4 @@
+import {expect} from 'vitest'
+import * as extensions from './dist/matchers'
+
+expect.extend(extensions)
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -16,7 +16,7 @@ jobs:
     if: ${{ !contains(github.head_ref, 'all-contributors') }}
     strategy:
       matrix:
-        node: [10.14, 12, 14, 15, 16]
+        node: [14, 16, 18, 20]
     runs-on: ubuntu-latest
     steps:
       - name: ⬇️ Checkout repo
--- a/extend-expect.js
+++ b/extend-expect.js
@@ -1,2 +1,0 @@
-// eslint-disable-next-line
-require('./dist/extend-expect')
--- a/jest-globals.d.ts
+++ b/jest-globals.d.ts
@@ -1,0 +1,1 @@
+/// <reference path="types/jest-globals.d.ts" />
--- a/matchers.d.ts
+++ b/matchers.d.ts
@@ -1,0 +1,3 @@
+import * as matchers from './types/matchers'
+
+export = matchers
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "0.0.0-semantically-released",
   "description": "Custom jest matchers to test the state of the DOM",
   "main": "dist/index.js",
+  "types": "types/index.d.ts",
   "engines": {
     "node": ">=8",
     "npm": ">=6",
@@ -19,8 +20,11 @@
   },
   "files": [
     "dist",
-    "extend-expect.js",
-    "matchers.js"
+    "types",
+    "*.d.ts",
+    "jest-globals.js",
+    "matchers.js",
+    "vitest.js"
   ],
   "keywords": [
     "testing",
@@ -32,7 +36,6 @@
   "license": "MIT",
   "dependencies": {
     "@babel/runtime": "^7.9.2",
-    "@types/testing-library__jest-dom": "^5.9.1",
     "aria-query": "^5.0.0",
     "chalk": "^3.0.0",
     "@adobe/css-tools": "^4.0.1",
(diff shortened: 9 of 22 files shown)


jest-globals.js
```diff
@@ -0,0 +1,4 @@
+const globals = require('@jest/globals')
```
jgoz, line 1: In my testing, `@jest/globals` had to use CommonJS

vitest.js
```diff
@@ -0,0 +1,4 @@
+import {expect} from 'vitest'
```
jgoz, line 1: And Vitest only supports ESM 🤷

vitest.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference path="types/vitest.d.ts" />
```
jgoz, line 1: A combination of `<reference>` and `export {}` in the referenced definitions seemed to be required to get TypeScript to augment the appropriate interfaces when this file was imported from node_modules.

README.md
```diff
@@ -144,6 +180,18 @@ haven't already:
   ],
 ```
 
+### With another Jest-compatible `expect`
```
kentcdodds, line 183: Would this be the approach if you're not using vitest's globals.

vitest.js
kentcdodds: In that case, should this file extension be .mjs?

Have you test this setup in a vitest project? If not, could you verify this will work as expected?

kentcdodds (commented):
Thank you so much for doing this work! I think it's going to be a very positive thing. I just have a couple questions.

vitest.js
jgoz: Great question. Using `.mjs` would be technically correct, but then all imported files would need to be ESM, which is not the case for matchers.js. Basically, vitest fails with "Error: Cannot find module".

I think we could change the extension after dual-publishing ES (#437?), but for now this works in vitest.

> Have you test this setup in a vitest project?

Yeah, I've got 4 test projects that I tested this PR against: Jest+JS, Jest+TS, @jest/globals+TS, vitest+TS. I'll publish them soon.

README.md
jgoz: I'm not 100% sure what you mean by vitest's globals. Vitest's default mode is `globals: false`, which requires explicit imports of describe, expect, etc., and the new vitest entrypoint is designed to work with that. I didn't actually think to test with `globals: true` and importing the `vitest/globals` types &mdash; I'll do that tomorrow.

This section was intended as a hedge against future testing libraries that might emerge that retain some level of compatibility with Jest. But if you think it might be more confusing than helpful, I'm happy to remove it.

jgoz: > I didn't actually think to test with `globals: true` and importing the `vitest/globals` types — I'll do that tomorrow.

Tested this and it works.

kentcdodds:
Thank you so much for helping with this!

MattyBalaam:
Very happy to see this 😊

TrevorBurnham:
Really excited to see this! FYI when I first plugged in this change (upgrading `@testing-library/jest-dom` from 5.16.5 to 6.0.0, upgrading `vitest` from 0.30.1 to 0.34.1 to fix a compatibility issue, then dropping `@types/testing-library__jest-dom` and adding `import "@testing-library/jest-dom/vitest"` to my test setup file), I got this error when running my tests:

```
Error: @vitejs/plugin-react can't detect preamble. Something is wrong. See https://github.com/vitejs/vite-plugin-react/pull/11#discussion_r430879201
```

No idea why that happened, but the error went away when I upgraded `@vitejs/plugin-react` from 3.1.0 to 4.0.4. Sharing this in hopes that it saves other devs some headaches. 🙏🏻

PR #513, opened 2023-08-16
fix: matchers type is making the global expect unsafe

kryops:
**What**:

Port of https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66247, after the types moved here in https://github.com/testing-library/jest-dom/pull/511

**Why**:

See https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/64858

**How**:

Moved the `Record<string, any>` constraint to the `matchers` export only, and removed it from the interface that augments the global Jest matchers.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged
      

I added a file with the TypeScript test cases from DefinitelyTyped, as they safeguard that the types should work correctly and prevent this issue from regressing. If they shouldn't be part of this repo, feel free to drop the last commit.


jgoz:
This is great, thank you for porting over the type tests. I'm going to test out these changes in my test repos tomorrow and then merge

jgoz (approved):
Thanks @kryops. I pushed a couple of commits to make this compatible with all 3 test platforms that we support and split/duplicated the type tests as such. It actually helped catch an issue with the `@jest/globals` interface definition.

Also enabled running the type tests in CI.

nickmccurdy:
@all-contributors add @kryops for bug code test

gnapse:
@all-contributors add @kryops for bug code test

PR #519, opened 2023-08-22
Publish ESM and CJS

### LINKED ISSUE #437: Publish ES modules

IanVS:
**What**:

This is a re-implementation of https://github.com/testing-library/jest-dom/pull/438 and https://github.com/testing-library/jest-dom/pull/453 which upgrades the `build` command to generate both esm and cjs bundles, for each of the entry files now exposed  in 6.0 (`index`, `matchers`, `jest-global`, and `vitest`).  

Closes https://github.com/testing-library/jest-dom/pull/498

**Why**:

Closes #437 

**How**:

I replaced kcd-scripts with rollup, configured to generate esm and cjs bundles.  

I added `exports` to the `package.json` to support these multiple exports, but lied slightly in the `types` field, since we only have one type definition for both cjs and esm, when technically we should have two, one for each format.  But, since we're not actually generating these definitions from typescript files, I think this should be fine, and we can adjust it in the future if/when the project is converted to TS itself.

I added files in `src` for `vitest` and `jest-global`, from which to build.  They are very similar to the files in the root, which will still be used by bundlers/typescript configurations which do not support package.json `exports`.

I tested this out in a typescript vitest project, using both `node` and `node16` moduleResolution.    I needed to add extensions to lodash imports to support Node16.

It would be excellent if @jgoz could review/test for jest-global support, since it's been a while since I've set up a jest project.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [X] Ready to be merged
      

I don't anticipate that this would require a breaking release, as it should continue to work as before.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,10 +1,10 @@
 const del = require('rollup-plugin-delete')
 
 const entries = [
-  'src/index.js',
-  'src/jest-globals.js',
-  'src/matchers.js',
-  'src/vitest.js',
+  './src/index.js',
+  './src/jest-globals.js',
+  './src/matchers.js',
+  './src/vitest.js',
 ]
 
 module.exports = [
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,4 +1,5 @@
 const del = require('rollup-plugin-delete')
+const commonjs = require('@rollup/plugin-commonjs')
 
 const entries = [
   './src/index.js',
@@ -26,6 +27,6 @@ module.exports = [
     ],
     external: id =>
       !id.startsWith('\0') && !id.startsWith('.') && !id.startsWith('/'),
-    plugins: [del({targets: 'dist/*'})],
+    plugins: [del({targets: 'dist/*'}), commonjs()],
   },
 ]
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,3 +1,5 @@
+const del = require('rollup-plugin-delete')
+
 const entries = [
   'src/index.js',
   'src/jest-globals.js',
@@ -24,5 +26,6 @@ module.exports = [
     ],
     external: id =>
       !id.startsWith('\0') && !id.startsWith('.') && !id.startsWith('/'),
+    plugins: [del({targets: 'dist/*'})],
   },
 ]
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,0 +1,28 @@
+const entries = [
+  'src/index.js',
+  'src/jest-globals.js',
+  'src/matchers.js',
+  'src/vitest.js',
+]
+
+module.exports = [
+  {
+    input: entries,
+    output: [
+      {
+        dir: 'dist',
+        entryFileNames: '[name].mjs',
+        chunkFileNames: '[name]-[hash].mjs',
+        format: 'esm',
+      },
+      {
+        dir: 'dist',
+        entryFileNames: '[name].js',
+        chunkFileNames: '[name]-[hash].js',
+        format: 'cjs',
+      },
+    ],
+    external: id =>
+      !id.startsWith('\0') && !id.startsWith('.') && !id.startsWith('/'),
+  },
+]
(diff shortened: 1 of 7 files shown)


rollup.config.js
```diff
@@ -0,0 +1,32 @@
...
+      },
+      {
+        dir: 'dist',
+        entryFileNames: '[name].js',
+        chunkFileNames: '[name]-[hash].js',
+        format: 'cjs',
+      },
+    ],
+    external: id =>
+      !id.startsWith('\0') && !id.startsWith('.') && !id.startsWith('/'),
```
jgoz, line 29: Out of curiosity, when would an import start with `\0`?

nickmccurdy:
Thanks for looking into this. Do you think it would be easy to upstream a generic solution for this into `kcd-scripts`? Then we could still have the fix here, but it should also be easier to use in other Testing Library package.

jgoz (commented):
I tested these changes locally with Jest, @jest/globals, and vitest with various TypeScript/module resolution settings and they all seemed to work. 🎉 

(I'm planning to move my example projects into this repo once I get back from vacation, which should help users and maintainers).

rollup.config.js
IanVS: It is a convention for rollup plugins that create virtual modules.  https://rollupjs.org/plugin-development/#conventions

We don't have any such plugins enabled here, so it could probably be removed if desired, but I kept it just to be safe.  Honestly I don't remember writing this originally, I may have copied it from somewhere else that suggested it.

jgoz: Thanks for the explanation. It's probably fine to keep in, I just haven't used rollup in several years and it stuck out to me. Esbuild has the convenient [`packages=external`](https://esbuild.github.io/api/#packages) option and Rollup doesn't appear to have anything similar.

jgoz (approved):
Thanks @IanVS! I have no opinion on upstreaming these changes into kcd-scripts, but publishing this as a canary makes sense.

jgoz:
Hmm, I assumed the `chore` scope would trigger a release, but apparently it does not...

jgoz:
Maybe someone else with push access could approve https://github.com/testing-library/jest-dom/pull/520 so these changes can get published?

nickmccurdy:
> > Thanks for looking into this. Do you think it would be easy to upstream a generic solution for this into kcd-scripts?
> 
> I'm not very familiar with kcd-scripts, but when I looked into it before, it relied on babel and didn't seem to be super-flexible in that regard. But I can spend some time looking into it. Though, it might be nice to publish this first, as a bit of a canary, then extend the approach to kcd-scripts after it's been battle tested by users a bit.

`kcd-scripts` is also based on Rollup, so this should be feasible. I wouldn't be against removing Babel, though it is possible to use them together.

PR #520, opened 2023-08-22
feat: manually trigger release

jgoz:
**What**:

Empty commit.

**Why**:

Release changes from #519 because `chore` doesn't auto-release.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged


IanVS:
Out of curiosity, how is the version determined (major/minor/patch)?  I'm guessing this should be released in a minor, not a patch, right?

nickmccurdy:
> Out of curiosity, how is the version determined (major/minor/patch)? I'm guessing this should be released in a minor, not a patch, right?

Ultimately it depends on [SemVer](https://semver.org/), though I'll try to help make a decision. Is any of the work in #519 a breaking change, or is it just meant to restore functionality from Jest DOM 15 in a back compatible way?

IanVS:
There should be no breaking changes, and it's not restoring any functionality.  It's publishing ESM modules for the first time, in addition to the CJS that was already being published previously.

jgoz:
Neither a breaking change nor restoring functionality. This is also not really functionality at all, it's just different packaging (dual-publishing to esm/cjs instead of just cjs).

IanVS:
Jinx.  ;)

nickmccurdy:
In that case I'd consider it SemVer minor, as it introduces a new API entry without affecting back compatibility.

jgoz:
@nickmccurdy Sounds good, changed to `feat`. Need another maintainer to approve now.

nickmccurdy (approved):

IanVS:
Thanks!  I guess it would be good to edit the release notes https://github.com/testing-library/jest-dom/releases/tag/v6.1.0 to include my PR that was not correctly titled (sorry, didn't know it needed a semantic title, maybe it would be good to add a check in CI for that).

PR #521, opened 2023-08-22
Update jest-globals.d.ts to match the @types/testing-library__jest-dom@5.14.9 definitions

heath-freenome:
**What**:

Add in the `global.jest` interface for the `Matchers` along side the `@jest/expect` in a manner similar to how it was defined in `@types/testing-library__jest-dom@5.14.9`.

**Why**:

The `@types/testing-library__jest-dom@5.14.9` defined both the  module for `expect` as well as the `global` for `jest`. When I switched to using the `@testing-library/jest-globals` from the old `@types` some of my use cases started failing with the following error:

```
error TS2339: Property 'toHaveTextContent' does not exist on type 'JestMatchers<HTMLElement>'.
```

**How**:

I copied the `global.jest` interface for `Matchers` from the `@types/testing-library__jest-dom@5.14.9` release and made a few changes so that it matched the current style/definition of the existing `@jest/expect`.

Here's the source from the [PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66389) that deleted it from `@types`

Screenshot 2023-08-22 at 12 45 42 PM

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged

### DIFF
--- a/types/jest-globals.d.ts
+++ b/types/jest-globals.d.ts
@@ -9,3 +9,13 @@ declare module '@jest/expect' {
       R
     > {}
 }
+
+declare global {
+  namespace jest {
+    interface Matchers<R = void | Promise<void>>
+      extends TestingLibraryMatchers<
+        typeof expect.stringContaining,
+        R
+      > {}
+  }
+}


jgoz (changes requested):
Hmm, this shouldn't be necessary. I think maybe there's some confusion here.

The `/jest-globals` entry point is intended to be used with `@jest/globals`, where you import `expect`, `it, etc., explicitly instead of relying on them being defined globally. The package is confusingly named because it's actually more like non-globals-mode.

If you are using Jest with its default options, you would not import from `/jest-globals`. You would import from the root:

```ts
import '@testing-library/jest-dom`;
```

jgoz:
@heath-freenome In version 6 we added multiple entry points for different use cases:

1. `@testing-library/jest-dom` &mdash; Jest + `@types/jest`
2. `@testing-library/jest-dom/jest-globals` &mdash; Jest with `injectGlobals: false`; tests import globals from `@jest/globals`
3. `@testing-library/jest-dom/vitest` &mdash; Vitest in either global or direct import mode

Your PR changes entry point 2 in the above list to make it behave like entry point 1, which is not correct because `expect` does not get added to the global namespace.

The main entry point (1) already has the code that you are proposing.

heath-freenome:
Closing because there is a better way

PR #523, opened 2023-08-23
fix(package.json): update main and module file paths

SteKoe:
**What**:
Updated *main* and *module* paths in package.json

**Why**:
I got the following error after upgrading to 6.1.0 in my project:

> SyntaxError: Named export 'parse' not found. The requested module '@adobe/css-tools' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

> import pkg from '@adobe/css-tools';
const { parse } = pkg;

After updating the paths in package.json and testing using `npm link` locally, everything works as expected, again.

**How**:

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,8 +2,8 @@
   "name": "@testing-library/jest-dom",
   "version": "0.0.0-semantically-released",
   "description": "Custom jest matchers to test the state of the DOM",
-  "main": "dist/cjs/index.js",
-  "module": "dist/esm/index.js",
+  "main": "dist/index.js",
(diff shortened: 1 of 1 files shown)


jgoz (approved):

nickmccurdy:
https://github.com/all-contributors add @SteKoe for bug code

nickmccurdy:
@all-contributors add @SteKoe for bug code

PR #525, opened 2023-08-23
fix: bump @adobe/css-tools for ESM support

### LINKED ISSUE #524: 6.1.0 @adobe/css-tools crashing
- `@testing-library/jest-dom` version: 6.1.0
- `node` version: 18
- `npm` (or `yarn`) version: yarn 3.6.0

### Relevant code or config:

n/a

### What you did:

Ran basic React tests, working fine with jest-dom@5, after updating to jest-dom@6 (this is important - NOT a clear install). 

### What happened:

```
SyntaxError: Named export 'parse' not found. The requested module '@adobe/css-tools' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from '@adobe/css-tools';
const { parse } = pkg;
```

### Problem description:

The required version of @adobe/css-tools jest-dom declares, ^4.0.1, is NOT correct. In my lockfile, it has already been resolved to 4.2.0. This hasn't changed with jest-dom@6 update, because ^4.0.1 is still the desired version range. This produced the error above. 

After investigation, it turned out @adobe/css-tools 4.3.0 is the first release with ESM support.

### Suggested solution:

Bump minimum @adobe/css-tools dependency to ^4.3.0.

jgoz:
**What**:

Fixes #524.

**Why**:

#519 added ESM support, but previous versions of `@adobe/css-tools` were not ESM compatible.

**How**:

Changed `package.json`

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
   "author": "Ernesto Garcia <gnapse@gmail.com> (http://gnapse.github.io)",
   "license": "MIT",
   "dependencies": {
-    "@adobe/css-tools": "^4.0.1",
+    "@adobe/css-tools": "^4.3.0",
     "@babel/runtime": "^7.9.2",
     "aria-query": "^5.0.0",
     "chalk": "^3.0.0",


jgoz:
@nickmccurdy

nickmccurdy (changes requested):
Blocked on a reproduction so I can test this (see https://github.com/testing-library/jest-dom/issues/524#issuecomment-1690587969).

wojtekmaj (approved):
Can confirm this fixes the issue.

@nickmccurdy It's going to be hard to automate this, because the problem is in the historical lockfile entries. You'll not encounter this problem when doing a fresh install without a lockfile. So you would need to "simulate the past" to do this. If you really want to however, here's how to do it (I'm providing instructions for Yarn 3 because that's what I use, but it'd be similar for any package manager):

* Install @testing-library/jest-dom & vitest (or Jest, but Vitest is natively ESM so it's faster to reproduce the issue)
* Find the following entry in the lockfile:
  ```yml
  "@adobe/css-tools@npm:^4.0.1":
    version: 4.3.1
    resolution: "@adobe/css-tools@npm:4.3.1"
    checksum: ad43456379ff391132aff687ece190cb23ea69395e23c9b96690eeabe2468da89a4aaf266e4f8b6eaab53db3d1064107ce0f63c3a974e864f4a04affc768da3f
    languageName: node
    linkType: hard
  ```
  Note that ^4.0.1 resolved to 4.3.1 _now_, but it must have resolved to an earlier version _in the past_. So to simulate this…
* Replace with:
  ```yml
  "@adobe/css-tools@npm:^4.0.1":
    version: 4.2.0
    resolution: "@adobe/css-tools@npm:4.2.0"
    checksum: dc5cc92ba3d562e7ffddb79d6d222c7e00b65f255fd2725b3d71490ff268844be322f917415d8c4ab39eca646343b632058db8bd5b1d646193fcc94d1d3e420b
    languageName: node
    linkType: hard
  ```
* Install again
* Observe the crash

frontendphil:
I've seen the same issue in our codebase (even though I also cannot share a reproduction case). However, when I added a `pnpm` override to use `@adobe/css-tools@4.3.0`, the particular error disappeared, but I got `TypeError: Cannot convert undefined or null to object` instead. So, the version bump might not be the ideal fix for this issue.

indooorsman:
in my project `@adobe/css-tools` was resolved and locked to `4.0.1` in past, and it does not have esm exports, as a result all my testings failed. uninstall it then reinstall it should work.

nickmccurdy (approved):
I can now confirm this is broken when using @jgoz's [reproduction](https://github.com/testing-library/jest-dom/issues/524#issuecomment-1691094563) and installing `@testing-library/jest-dom`, but fixed when linking the built package with this PR. Thanks!

PR #552, opened 2023-11-28
fix: support uppercase custom props in toHaveStyle

### LINKED ISSUE #551: toHaveStyle fails on custom property names containing uppercase letters
The error will be the dreaded _Compared values have no visual difference._

### Relevant code or config:

It's easy to reproduce by changing the existing test:

https://github.com/testing-library/jest-dom/blob/4ae0231be83f3ca79ae3ff735cf6c0ceb399fea4/src/__tests__/to-have-style.js#L122-L128

```js
  test('handles inline custom properties', () => {
    const {queryByTestId} = render(`
      <span data-testid="color-example" style="--backgroundColor: blue">Hello World</span>
    `)
    expect(queryByTestId('color-example')).toHaveStyle('--backgroundColor: blue') // this will fail
  })
```

The culprit is the `.toLowerCase()` here:

https://github.com/testing-library/jest-dom/blob/4ae0231be83f3ca79ae3ff735cf6c0ceb399fea4/src/to-have-style.js#L23

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties#sect2):

> **Note**: Custom property names are case sensitive — `--my-color` will be treated as a separate custom property to `--My-color`.

Current workarounds are changing the code under test, including the stylesheets, to use all-lowercase custom property names, or to not use `.toHaveStyle()`.

depoulo:
**What**:

support uppercase custom props in toHaveStyle

fixes #551

**Why**:

CSS custom properties are case sensitive

**How**:

don't lowercase custom property names before comparing them

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged
      

--

please squash and merge

related: #148, #164

### DIFF
--- a/src/__tests__/to-have-style.js
+++ b/src/__tests__/to-have-style.js
@@ -121,9 +121,9 @@ describe('.toHaveStyle', () => {
 
   test('handles inline custom properties', () => {
     const {queryByTestId} = render(`
-      <span data-testid="color-example" style="--color: blue">Hello World</span>
+      <span data-testid="color-example" style="--accentColor: blue">Hello World</span>
     `)
-    expect(queryByTestId('color-example')).toHaveStyle('--color: blue')
+    expect(queryByTestId('color-example')).toHaveStyle('--accentColor: blue')
   })
 
   test('handles global custom properties', () => {
--- a/src/__tests__/to-have-style.js
+++ b/src/__tests__/to-have-style.js
@@ -70,6 +70,7 @@ describe('.toHaveStyle', () => {
       background-color: black;
       color: white;
       float: left;
+      --var-name: 0px;
       transition: opacity 0.2s ease-out, top 0.3s cubic-bezier(1.175, 0.885, 0.32, 1.275);
     }
   `
@@ -90,6 +91,11 @@ describe('.toHaveStyle', () => {
       expect(container.querySelector('.label')).toHaveStyle(
         'transition: all 0.7s ease, width 1.0s cubic-bezier(3, 4, 5, 6);',
       ),
+    ).toThrowError()
+
+    // Custom property names are case sensitive
+    expect(() =>
+      expect(container.querySelector('.label')).toHaveStyle('--VAR-NAME: 0px;'),
     ).toThrowError()
 
     // Make sure the test fails if the css syntax is not valid
--- a/src/__tests__/to-have-style.js
+++ b/src/__tests__/to-have-style.js
@@ -119,7 +119,7 @@ describe('.toHaveStyle', () => {
     )
   })
 
-  test('handles inline custom properties', () => {
+  test('handles inline custom properties (with uppercase letters)', () => {
     const {queryByTestId} = render(`
       <span data-testid="color-example" style="--accentColor: blue">Hello World</span>
     `)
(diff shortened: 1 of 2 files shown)


gnapse (approved):

src/to-have-style.js
```diff
@@ -20,7 +20,7 @@ function isSubset(styles, computedStyle) {
     Object.entries(styles).every(
       ([prop, value]) =>
         computedStyle[prop] === value ||
-        computedStyle.getPropertyValue(prop.toLowerCase()) === value,
+        computedStyle.getPropertyValue(prop) === value,
```
gnapse, line 23: Wait, I think we may have a problem. I suspect that it was done this way because non-custom properties are case-insensitive. At least a quick check in the browser shows that setting `Display: none` hides the element.

Can you make it so that we keep the lower case transformation when the prop does not look like a custom css prop?

gnapse (changes requested):

src/to-have-style.js
depoulo: Interesting, I'll look into it tomorrow!

depoulo: Good that there was a test case added with #148. Could have found out via `git blame`.

src/__tests__/to-have-style.js
```diff
@@ -214,7 +214,7 @@ describe('.toHaveStyle', () => {
         <span data-testid="color-example" style="font-size: 12rem">Hello World</span>
       `)
       expect(() => {
-        expect(queryByTestId('color-example')).toHaveStyle({ fontSize: '12px' })
+        expect(queryByTestId('color-example')).toHaveStyle({fontSize: '12px'})
```
depoulo, line 217: That one came from the husky pre-commit hook formatting the whole file. By the way running `npm run format` would have changed a bunch of other files, too.

src/to-have-style.js
```diff
@@ -20,6 +20,7 @@ function isSubset(styles, computedStyle) {
     Object.entries(styles).every(
       ([prop, value]) =>
         computedStyle[prop] === value ||
+        computedStyle.getPropertyValue(prop) === value ||
         computedStyle.getPropertyValue(prop.toLowerCase()) === value,
```
depoulo, line 24: > Can you make it so that we keep the lower case transformation when the prop does not look like a custom css prop?

After reading the whole discussion [here](https://github.com/testing-library/jest-dom/pull/154#discussion_r345172964), I thought this would be the least disruptive way.

gnapse: I still think that we may need to detect specifically if the prop name corresponds to a custom CSS property. They are case-insensitive. The way the code is right now, if I have an element with `--var-name: 0px`, a test checking for `--VAR-NAME: 0px` would pass, and that's not accurate.

Can't you do something along the lines of:

```js
const isCustomProp = prop.startsWith('--')
if (!isCustomProp) {
  prop = prop.toLowerCase()
}
// Keep the condition as before
```

gnapse (changes requested):

src/to-have-style.js
depoulo: see my latest commit

gnapse (approved):

PR #555, opened 2023-12-05
fix: upgrade @adobe/css-tools to 4.3.2 to address vulnerability

lernerb:
**What**:
Per GitHub Advisory (https://github.com/advisories/GHSA-prr3-c3m5-p7q2), updating to 4.3.2 for all consumers of this package.

**Why**:
All consumers would have to do an `audit fix` and manually override to 4.3.2, this forces it for folks.

**How**:
Updates the package.json file to force one higher patch version that does not have the security issue.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
   "author": "Ernesto Garcia <gnapse@gmail.com> (http://gnapse.github.io)",
   "license": "MIT",
   "dependencies": {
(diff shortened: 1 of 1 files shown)


lernerb:
Ah looks like someone beat me to this, i'll close this in favor of https://github.com/testing-library/jest-dom/pull/553

PR #563, opened 2023-12-26
feat: Support for regular expressions in toHaveClass

### LINKED ISSUE #556: Regex support for `.toHaveClass()`

With CSS-in-JS and CSS modules, it's pretty common for CSS classes to contain randomly generated unique strings appended to the end. This makes `.toHaveClass()` extremely limited.

While there are workarounds involving string matching on the actual class string, these are not ideal because many projects prefer to leave this eslint rule turned on: https://github.com/testing-library/eslint-plugin-jest-dom/blob/main/docs/rules/prefer-to-have-class.md

If `.toHaveClass()` could accept RegExp in addition to strings, that would allow projects to continue using [eslint: prefer-to-have-class](https://github.com/testing-library/eslint-plugin-jest-dom/blob/main/docs/rules/prefer-to-have-class.md) while also using CSS Modules / CSS-in-JS.

### Suggested implementation:

A backwards compatible implementation that simply adds support for RegExp matching for `.toHaveClass()`.

```
expect(deleteButton).toHaveClass(/^text-sm/)
expect(deleteButton).not.toHaveClass(/^text-lg/)
```

gnapse:
**What**:

Enhances `.toHaveClass` to also support receiving regular expressions in its params, to be matched against the target element's class names.

Since this is a newly supported argument type, this is not a breaking change. Any code calling `.toHaveClass` with strings only (the only supported use so far) that code will continue to work as before.

**Why**:

Closes #556 

It enables more freedom when checking an element's class names. For instance, in some preprocessed CSS scenarios, class names are generated programmatically, and the writer of the test cannot know in advance the exact class name, though they may know a substring or pattern in it.

**How**:

By checking, in a handful of places of the relevant code, whether a value is a string or a regular expression. If it is a regular expression, then we do a different check.

Also, in the particular case of passing `{exact:true}` options to `.toHaveClass`, we disable regular expressions support, and throw an error if regular expressions are found in the arguments. 

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged


eps1lon (approved):
looks good

PR #566, opened 2024-01-04
fix: Standalone types for "./matchers" export and add Bun support

### LINKED ISSUE #554: TypeScript types for the `"./matchers"` export don't match implementations
- `@testing-library/jest-dom` version: 6.1.5
- `node` version: N/A - using Bun 1.0.15
- `jest` (or `vitest`) version: N/A - using `bun test`
- `npm` (or `yarn`) version: N/A - using Bun 1.0.15
- `@testing-library/react` version: 14.1.2

### Relevant code or config:

This code works in practice, it's just the types that are causing a problem:

```ts
import * as matchers from "@testing-library/jest-dom/matchers";
import { expect, mock, test } from "bun:test";

expect.extend(matchers);
//            ^ Argument of type 'TestingLibraryMatchers<any, void> & Record<string, any>' is not assignable to parameter of type 'ExpectExtendMatchers...
```

### What you did:

Tried to extend `expect` from `bun:test` with matchers from `@testing-library/jest-dom/matchers`.

I don't think this is specific to Bun, but they just added support for `expect.extend` so I thought I'd give it go.

### What happened:

The TypeScript definitions for the `"./matchers"` export don't seem to match the implementations of the actual matchers.

Take `toBeChecked`, for example. The implementation is here:

https://github.com/testing-library/jest-dom/blob/b7b7c6a9652f259434d13a22e4319826a4bd4d8b/src/to-be-checked.js#L4

But the type definition doesn't include the `element` parameter:

https://github.com/testing-library/jest-dom/blob/b7b7c6a9652f259434d13a22e4319826a4bd4d8b/types/matchers.d.ts#L433

Also, the `R` generic is specified as `void` in the `export` statement, so the return type is effectively missing as well. (Should be something like `{ pass: boolean; message: () => string; }`.)

https://github.com/testing-library/jest-dom/blob/b7b7c6a9652f259434d13a22e4319826a4bd4d8b/types/matchers.d.ts#L667-L669

### Reproduction:

See "Relevant code" section above.

### Problem description:

Types should match the implementation.

### Suggested solution:

https://github.com/testing-library/jest-dom/blob/main/types/matchers.d.ts should be updated to match the implementation of the matcher functions. If those types are correct in some other context (since they do seem to match the implementation when called as a property of an `expect()` result), then a separate definition file should be created for the `"./matchers"` export.

jakeboone02:
**What**:

Fixes #554, type definitions for "./matchers" export.

Also adds support for Bun types when using `bun:test`.

**Why**:

This results in a TypeScript error:

```
import * as matchers from "@testing-library/jest-dom/matchers"
export.extend(matchers) // <-- matchers is not the correct type
```

Also...Bun is great.

**How**:

Referenced a new `matchers-standalone.d.ts` file in package.json#export#"./matchers" (not exported in the main export) that contains a version of `matchers.d.ts` converted to the correct types.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Ready to be merged

I don't think new documentation is necessary, but I'm not sure this is ready to be merged because I had to disable the husky config in order to commit. The part of the pre-commit script that failed was `kcd-scripts test --findRelatedTests`, I think because it saw a new file that didn't contain tests. But the new file doesn't really require JavaScript tests since it's a `.d.ts` file. Also the new Bun files have tests, so I'm not really sure what the husky config's problem is.

**Edit 1:** To clarify, this PR does _not_ change the types for the main `"."` export. It only fixes the types for the `"./matchers"` export.

**Edit 2:** Looks like the husky/commit thing wasn't an issue and this is actually ready to be merged.


gnapse (approved):
I think it looks good, but I'll wait for one or two other reviews from our maintainers before merging.

Also, does it make sense to keep `matchers.d.ts` if you are replacing its export with the new `matchers-standalone.d.ts` file?

jakeboone02:
Yes, it makes sense because this new code is not replacing the matchers types in the main export.  This PR only fixes the types for the "./matchers" export.

jakeboone02:
@gnapse Do you know if any other maintainers have looked at this? It's a relatively minor change, and as Bun (`bun:test` in particular) becomes more popular, this will avoid a lot of bug reports.

gnapse:
It does not seem like they've seen it, but let's see if we can go ahead with this.

One last question before we merge: are you confident that this is a patch release? On one hand, I'm wondering if it should be a minor release, as it adds a new feature (bun support). OTOH, I wonder if it isn't a breaking change even. It does not look like it, but I wanted to make sure first.

jakeboone02:
> It does not seem like they've seen it, but let's see if we can go ahead with this.
> 
> One last question before we merge: are you confident that this is a patch release? On one hand, I'm wondering if it should be a minor release, as it adds a new feature (bun support). OTOH, I wonder if it isn't a breaking change even. It does not look like it, but I wanted to make sure first.

I would say it's a judgement call whether this is a patch or minor. You could argue that the addition of Bun support is "fixing a bug specific to Bun," but it does seem like a new feature in a way, so strictly speaking a minor might be more appropriate per semver.

I'm all but certain this is not a breaking change. The only reason I'm not _completely_ certain is there are just too many different environments and configurations in JS/TS land to be 100% certain of anything.

axeleriksson147:
Hello @jakeboone02 and @gnapse 

This breaks the `TestingLibraryMatchers` import for us in TypeScript. We have a file something like:

```
import { expect as _expect } from 'expect';
import { TestingLibraryMatchers } from '@testing-library/jest-dom/matchers';

declare module 'expect' {
  export interface Matchers<R = void> extends TestingLibraryMatchers<typeof _expect.stringContaining, R> {}
}

declare global {
  const expect: typeof _expect;
}
```

This way importing expect was including the extended TestingLibraryMatchers. With version `6.2.1` the import is incorrect.

jakeboone02:
@axeleriksson147 is the `TestingLibraryMatchers` type not available from the main `"@testing-library/jest-dom"` export? If you put a minimal repo together I'll take another look.

gnapse:
Thanks for taking a look @jakeboone02. Please keep me in the loop if something actionable comes from it (e.g. if we have something to change and release to make things right again).

AxelEriksson0:
@gnapse and @jakeboone02 I created a minimal repo - https://github.com/AxelEriksson0/testing-library-jest-dom-6.2.0-breaking-type-location

My suggestion is to just export `TestingLibraryMatchers` in the matchers.standalone.ts.

jakeboone02:
> @gnapse and @jakeboone02 I created a minimal repo - https://github.com/AxelEriksson0/testing-library-jest-dom-6.2.0-breaking-type-location
> 
> My suggestion is to just export `TestingLibraryMatchers` in the matchers.standalone.ts.

Thanks @AxelEriksson0! I [patched the "broken" project with your suggestion](https://github.com/jakeboone02/testing-library-jest-dom-6.2.0-breaking-type-location/commit/91bd0d9e7425a1f4bd18c9b24ed9b30717e88781) and it seems to have done the trick.

If you think that resolves the issue I'll submit another PR.

axeleriksson147:
> > @gnapse and @jakeboone02 I created a minimal repo - https://github.com/AxelEriksson0/testing-library-jest-dom-6.2.0-breaking-type-location
> > My suggestion is to just export `TestingLibraryMatchers` in the matchers.standalone.ts.
> 
> Thanks @AxelEriksson0! I [patched the "broken" project with your suggestion](https://github.com/jakeboone02/testing-library-jest-dom-6.2.0-breaking-type-location/commit/91bd0d9e7425a1f4bd18c9b24ed9b30717e88781) and it seems to have done the trick.
> 
> If you think that resolves the issue I'll submit another PR.

Yeah, I think so. I just don't see `TestingLibraryMatchers` being exposed anymore, unless I'm missing something. But that should resolve it, thank you!

paulleonartcalvo:
@jakeboone02 Just came across this thread and am having what appears to be a unique issue. Im extending `bun`'s expect with these matchers, but on error, some of the matchers like `toBeInTheDocument` cause errors related to the jest matcher utils like the following:

```
TypeError: this.utils.RECEIVED_COLOR is not a function. (In 'this.utils.RECEIVED_COLOR(this.isNot ? errorFound() : errorNotFound())', 'this.utils.RECEIVED_COLOR' is undefined)
```

Is this something that has been encountered or that a solution exists for?

jakeboone02:
@paulleonartcalvo can you provide a reproduction?

paulleonartcalvo:
@jakeboone02 👍  I'll try and set up an environment tonight. For awareness, i was using Bun + typescript + React + Vite

Issue #143, opened 2019-10-17
Add toHaveRole


babramczyk:
### Describe the feature you'd like:

A `toHaveRole` matcher that uses a similar implementation to `@testing-library`'s described [here](https://testing-library.com/docs/dom-testing-library/api-queries#byrole). In other words, checking implicit roles in addition to `aria` roles (i.e. using `aria-query`).

### Motivation

The use case that prompted this was a page that had a button (example text "Create Issue") that was supposed to navigate to a page with a heading with the same text ("Create Issue").

Ideally, the test would look like this:

```javascript
const createIssueBtn = getByText(/create issue/i)
expect(createIssueBtn).toHaveRole('button')
fireEvent.click(createIssueBtn)

const createIssueHeader = getByText(/create issue/i)
expect(createIssueHeader).toHaveRole('header')
```

Instead, the opposite can only be done (i.e. `getByRole`, then assert with `toHaveTextContent`). This is a little more fickle, since it relies on these elements being the first of the specified role on the page, which may not always be the case. Querying by text first is much more specific, and hopefully more robust.

Concretely, here's what I had to implement as an alternative:

```javascript
    const utils = render(<App />)
    const createIssueBtn = utils.getByText(/create issue/i)
    // Since the heading on the page we're going to also has the text "create issue", we need to be a little dirty and make sure this element is a button
    expect(createIssueBtn.tagName).toBe('BUTTON')

    fireEvent.click(createIssueButton)

    // Wait for the "create issue" element to be a heading instead of a button -- necessary since the click won't trigger the page change immediately
    await wait(() =>
      expect(utils.getByRole('heading')).toHaveTextContent(/create issue/i)
    )
    const createIssueHeader = utils.getByText(/create issue/i)
    // Again, just being paranoid and making sure this element isn't just the button that's on the home page
    expect(examSchedulerHeader.tagName).not.toBe('BUTTON')
```

connormeredith:
It feels like it might be simpler to assert that the user navigated to the correct URL after clicking button, getting the history object in a similar way to: https://testing-library.com/docs/example-react-router#reducing-boilerplate

```jsx
const utils = render(<App />)

const createIssueBtn = utils.getByText(/create issue/i)
fireEvent.click(createIssueBtn)

expect(history.location.pathname).toBe('/create-issue-page');
```

Or would it be possible to use [`waitForDomChange`](https://testing-library.com/docs/dom-testing-library/api-async#waitfordomchange) after you click the button?

[...]

PR #572, opened 2024-01-28
feat: Add toHaveRole matcher

fpapado:
**What**:

This PR adds the `toHaveRole` matcher described in #143.

**Why**:

Sometimes, it is desirable to make assertions about the accessible role of an element. While the [prioritites in testing-library](https://testing-library.com/docs/queries/about/#priority) encourage querying an element by its role, there are times when you already have access to an element, and want to run assertions, without changing the query itself. For example, some codebases prioritise test ids over role queries. This enables more robust accessibility checks for these scenarios :)

**How**:

In large part, I am porting over [dom-testing-library's role helpers](https://github.com/testing-library/dom-testing-library/blob/bd04cf95a1ed85a2238f7dfc1a77d5d16b4f59dc/src/role-helpers.js#L80) to this one. 

Under the hood, the role helpers use [aria-query](https://github.com/A11yance/aria-query/tree/main) which gives programmatic (and auto-updating) access to the ARIA specification and its HTML mappings.

The helpers then compile the different conditions (attributes, DOM tree relationships) into a CSS selector, and provide a function to match an element via the `HTMLElement` `matches()` method.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged
      

## Questions for the maintainers

I have two main questions, that I would want the maintainers to clarify:
1. The current approach relies on using CSS selectors. Are they supported in all the environments that this library supports? Porting the selectors over to attribute checks is doable, in a pretty automated way (we'd change the implementation of processing `aria-query` mappings)
2. Since we are copying over the function from `dom-testing-library`, I have a gut feeling that we could move these helpers to a shared package. I noticed that @eps1lon's [dom-accessibility-api](https://github.com/eps1lon/dom-accessibility-api) also has a [`getRole function`](https://github.com/eps1lon/dom-accessibility-api/blob/main/sources/getRole.ts), albeit with a different implementation. I have not looked enough into it, to evaluate which one is more fit for this use-case. Do you have any preferences?

Question 1. is more imminent for this PR, I suppose 😁 Question 2. can be an implementation detail, that we can tackle in a follow-up. I leave it up to you.

Please let me know if there is any other information that I can provide!


eps1lon:
> Since we are copying over the function from dom-testing-library, I have a gut feeling that we could move these helpers to a shared package.

I wouldn't worry about this part too much. dom-accessibility-api is the natural fit but we don't want to use aria-query in there so it's not just about copying over impl but getting rid of aria-query entirely.

eps1lon:
Did you file a docs PR yet?

fpapado:
> Did you file a docs PR yet?

I made the changes to README.md in this PR. Is there some other place to change for the docs, or should I split the docs and feature into two PRs?

gnapse (approved):

fpapado:
Thank you 😌 

> However, I do sympathize with the idea of the shared implementation. I wonder why `dom-testing-library` does not use the one from `dom-accessibility-api`. Perhaps that is a change that can be contributed to that library.

My gut feeling is that `dom-testing-library` prioritises _querying_ for roles, so compiling the rules to a CSS selector might be better-suited. For a matcher, where we already have a handle on the relevant element, the approach in `dom-accessibility-api` seems equally viable. For me personally, it's mostly a question of where the libraries pull their sources from, and what the update cadence is. It's tricky to keep up with the mapping specs (they change slowly, but also in subtle ways).

For invalid `role`, I believe browsers preserve it, even if the value is invalid. I guess, in a way, this enables the fallbacks and future roles to exist. For example, here is Chrome's accessibility devtools with this [codepen of role="whatever"](https://codepen.io/fpapado/pen/rNRppgZ):

CleanShot 2024-01-29 at 13 09 10@2x

So I can see some value in not being strict about them, but I agree that matchers can make different decisions than the browser 😅

eps1lon:
>   I wonder why dom-testing-library does not use the one from dom-accessibility-api.

Historicaly reasons. A rewrite would not use aria-query. Ideally we'd move everything into dom-accessibility-api (the interface not the implementation) and use that everywhere.

gnapse:
@all-contributors, please add @fpapado for code, docs, test

gnapse:
@all-contributors please add @fpapado for code, docs, test

Issue #574, opened 2024-01-30
Types breakage in v6.4.0


Rugvip:
We're seeing an issue with the types in the latest release (v6.4.0). In particular the addition of the [`ByRoleMatcher` type declaration](https://github.com/testing-library/jest-dom/blob/f7dc673dbc13e508f9867a9eb9c547ce3964b22c/types/matchers.d.ts#L5) breaks the [export assignment](https://github.com/testing-library/jest-dom/blob/f7dc673dbc13e508f9867a9eb9c547ce3964b22c/types/matchers.d.ts#L713) later on, resulting in the following error:

```
TS2309: An export assignment cannot be used in a module with other exported elements.
```

- `@testing-library/jest-dom` version: v6.4.0
- `node` version: n/a
- `jest` (or `vitest`) version: n/a
- `npm` (or `yarn`) version: n/a

### Relevant code or config:

[tsconfig.json](https://github.com/backstage/backstage/blob/master/packages/cli/config/tsconfig.json)

### What you did:

Example breakage in a CI run here: https://github.com/backstage/backstage/actions/runs/7708717042/job/21008397598?pr=22596

### What happened:

```
node_modules/@testing-library/jest-dom/types/matchers.d.ts:713:1 - error TS2309: An export assignment cannot be used in a module with other exported elements.

713 export = matchers
    ~~~~~~~~~~~~~~~~~


Found 1 error in node_modules/@testing-library/jest-dom/types/matchers.d.ts:713
```

### Reproduction:

Doesn't seem to reproduce in codesandbox, may be related to TypeScript version or config?

### Problem description:

Breaks TypeScript type checks.

### Suggested solution:

Might be enough to switch the re-export to `export { matchers }`?

fpapado:
Hi @Rugvip! I am not a maintainer, but I am the person that contributed `.toHaveRole`, so I take some of the responsibility 😌 First of all, my apologies for the hassle!

## Two quick checks

Can I ask you to check two things on your end?

1. If you modify `matchers.d.ts` to remove the `export` declaration for `ByRoleMatcher`, are the type errors fixed?
2. If not, can you check removing the `import` and `export` for `ByRoleMatcher`, and changing this line to [`role: string`](https://github.com/testing-library/jest-dom/blob/f7dc673dbc13e508f9867a9eb9c547ce3964b22c/types/matchers.d.ts#L624)?

[...]

PR #575, opened 2024-01-30
fix: Remove errant export of GetByRoleMatcher, fixing type checking in some TS configurations

fpapado:
**What**:

Follow-up to #143 , and the issue reported at #574.

Mixing `export type` with `export = matchers` can throw off TypeScript in some configurations. We do not need to export the matcher as a separate type, since the argument types are available in the matcher already.

I will admit, I do not know the exact combination of configuration that causes this issue. I think **it would be desirable to include a test case for it**, to ensure that we do not break it accidentally in the future. 

The types are doing quite a lot of heavy lifting (types for explicit extend, types on globals/declaration merging, plus being "external" to the source), so perhaps we are not testing all the combinations. I am a bit surprised that we did not get this error in the existing tests, the double export sounds conceptually invalid. The issue at #574 has a reproduction, but paring it down to the size of the existing tests might be best.

Anyway, in terms of compatibility, I am confident that this is the best approach. By avoiding the extra `export`, we do not confuse TypeScript, and we are back in the previous status quo (`exports = matchers` only).

**Why**:

Avoiding unintended type changes, in some environments.

**How**:

By reverting what the TypeScript error was pointing at:
> TS2309: An export assignment cannot be used in a module with other exported elements.

**Checklist**:

- [X] Documentation
- [ ] Tests
- [X] Updated Type Definitions
- [X] Ready to be merged
      

I think this is ready to be merged, but the lack of an inverse test case is bothering me 😅 

Please let me know if I can help in any way, and I'll give it a shot!


types/matchers.d.ts
```diff
@@ -1,9 +1,5 @@
 import {ARIARole} from 'aria-query'
```
andykenward, line 1: Could this be changed to? 
```typescript
import {type ARIARole} from 'aria-query'
```

fpapado:
Follow-up: we found in https://github.com/testing-library/jest-dom/issues/574 that `skipLibCheck` is what causes this error to silently get ignored from the tests. I don't have a concrete suggestion for what to do with that; perhaps a separate tsconfig without `skipLibCheck`, but that would need to account for some library type errors that crop up.

Should we merge this to fix current breakage? Gentle ping @gnapse 😌

gnapse (approved):

PR #582, opened 2024-02-16
fix: toHaveStyle assertion with invalid style (#564)

### LINKED ISSUE #564: toHaveStyle for property `font-size` behaves incorrectly for number values
- `@testing-library/jest-dom` version: 6.1.5
- `node` version: 20.3.1
- `jest` (or `vitest`) version: 0.34.1
- `npm` (or `yarn`) version: 9.6.7

### Relevant code or config:

```js
render(<div data-testid="element" style={{ fontSize: 8 }} />)
expect(screen.getByTestId('element')).toHaveStyle({ fontSize: 1 })
```

### What you did:
I expected the above assertion to fail

### What happened:
The test passes instead!✅

### Problem description:
the assertion is clearly incorrect and can easily cause bugs

### Suggested solution:
I have not thought of a solution yet but would like to investigate this issue during my off hours.

EduardoSimon:
**What**:

Given an invalid declaration such as `fontSize: 8`, due to the missing unit, the `toHaveStyle` matcher should not pass the following test:

```
render(<div data-testid="element" style={{ fontSize: 8 }} />)
expect(screen.getByTestId('element')).toHaveStyle({ fontSize: 1 })
```

**Why**:

Refer to the original issue for more info: #564 

**How**:

This PR fixes #564 by adding a more restrictive guard in the matcher's logic.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/to-have-style.js
+++ b/src/to-have-style.js
@@ -14,6 +14,14 @@ function getStyleDeclaration(document, css) {
   return styles
 }
 
+function isInvalidStyleDeclaration(name, value, computedStyle) {
+  return (
+    name &&
+    !value &&
+    !computedStyle[name] &&
+    !computedStyle.getPropertyValue(name)
+  )
+}
 function isSubset(styles, computedStyle) {
   return (
     !!Object.keys(styles).length &&
@@ -22,11 +30,16 @@ function isSubset(styles, computedStyle) {
       const spellingVariants = [prop]
       if (!isCustomProperty) spellingVariants.push(prop.toLowerCase())
 
-      return spellingVariants.some(
-        name =>
+      return spellingVariants.some(name => {
+        if (isInvalidStyleDeclaration(name, value, computedStyle)) {
+          return false
+        }
+
+        return (
           computedStyle[name] === value ||
-          computedStyle.getPropertyValue(name) === value,
-      )
+          computedStyle.getPropertyValue(name) === value
+        )
+      })
     })
   )
 }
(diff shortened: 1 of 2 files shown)


src/to-have-style.js
```diff
@@ -14,6 +14,14 @@ function getStyleDeclaration(document, css) {
...
 }
 
+function isInvalidStyleDeclaration(name, value, computedStyle) {
+  return (
+    name &&
+    !value &&
+    !computedStyle[name] &&
+    !computedStyle.getPropertyValue(name)
+  )
+}
```
gnapse, line 24: nit: keep blank line between function declarations:

```suggestion
}

```

gnapse, line 24: It is not fully evident (to me at least) why this works. I'm not saying I doubt that it works, I'm saying that perhaps this function could use a jsdoc to explain what it does. Or at the very least can you share here the logic? Perhaps no comment is needed, and I'm just having trouble to understand it myself.

gnapse (approved):
Looks good, thanks!

Although it may benefit from a comment on the new helper function as mentioned/requested below.

src/to-have-style.js
EduardoSimon: Good point! Actually getting it to work was a bit tricky, so a comment would not harm.

EduardoSimon: Thanks for the suggestion, have you thought of adding a pre-commit hook to automatically apply the library's code style automatically?

EduardoSimon:
Hey @gnapse , is it possible to add the Hacktoberfest label to the PR?

Issue #515, opened 2023-08-17
Upgrading to v6.0.0 remove the `toBeInTheDocument` interface with vitest


qchuchu:
- `@testing-library/jest-dom` version: 6.0.0
- `node` version: 18.16
- `npm` (or `yarn`) version: 9.5.1

### Relevant code or config:

```ts
import '@testing-library/jest-dom/vitest';
import { vitest } from 'vitest';
```

```ts
  await waitFor(() => {
    expect(getByTestId('text-input-error-message')).toBeInTheDocument();
  });
```

### What you did:

Upgraded from v5 to v6

### What happened:

<img width="654" alt="image" src="https://github.com/testing-library/jest-dom/assets/45853629/8972b20b-de1a-46a5-a2b3-267cbd55322f">

I don't know why, but upgrading the package to the last version using vitest caused all my "toBeInTheDocument" matchers to be undefined, although I can see them in the type definition of the package.

### Reproduction:

TO-DO

### Problem description:

- All my tests are failing because toBeInTheDocument is not defined anymore

### Suggested solution:

I guess this is due to an upgrade on the dependency of jest types but I'm not sure. Still digging.

jgoz:
What version of vitest, and where is the import statement located for @testing-library/jest-dom/vitest?

curtvict:
fwiw this is happening to me in Jest too.

qchuchu:
ah sorry I thought that I added the vitest version.

Vitest -> 0.34.2

I added the import statement in the vitest.setup.ts file (that is pointed from my vite.config.ts like this) 

```ts
/// <reference types="vitest" />

import react from '@vitejs/plugin-react-swc';
import { defineConfig } from 'vite';

// https://vitejs.dev/config/
export default defineConfig({
  plugins: [react()],
  test: {
    globals: true,
    environment: 'jsdom',
    setupFiles: ['test/vitest.setup.ts'],
    include: ['**/?(*.)test.ts?(x)'],
  },
});
```

qchuchu:
So it does work by adding the setup file in the tsconfig.json, but only if I stay with the original configuration

`test/vitest.setup.ts`
```ts
import '@testing-library/jest-dom';
```

If I add /vitest at the end, i'm facing the issue anymore.

jgoz:
Hmm, I think that's because you're using Vitest in "globals" mode. There isn't really a difference between what you have now and using `/vitest`, which was designed for the default non-globals mode, so if this works for you I say go with it.

[...]

PR #589, opened 2024-03-11
fix: proper vitest ts support (#515)

nyaapass:
**What**:

This PR [extends Vitest's default `Assertion` interface](https://vitest.dev/guide/extending-matchers.html#:~:text=If%20you%20are%20using%20TypeScript%2C%20since%20Vitest%200.31.0%20you%20can%20extend%20default%20Assertion%20interface%20in%20an%20ambient%20declaration%20file%20(e.g%3A%20vitest.d.ts)%20with%20the%20code%20below%3A) to fix the `toBeInTheDocument` ts error when using Vitest (#515).

**Why**:

I don't know (still cannot reproduce #515 on my own computer). But [from this comment](https://github.com/testing-library/jest-dom/issues/515#issuecomment-1987823474), adding 

```ts
/// <reference types="@testing-library/jest-dom" />
``` 

in custom `vitest-setup.ts` file would fix this error in that situation. Therefore I think the `types/vitest.d.ts` may need to be updated.

And when I install this repo's dependencies by pnpm and running `tsc -p types/__tests__/vitest`, there would be similar errors of #515 (if using npm or `import '../../jest'` instead of `import '../../vitest'` like that comment, there would be no error). This PR would fix this error even using pnpm.

image

**How**:

https://vitest.dev/guide/extending-matchers.html

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/types/vitest.d.ts
+++ b/types/vitest.d.ts
@@ -1,6 +1,10 @@
 /* eslint-disable @typescript-eslint/no-shadow */
 
-import {type Assertion, type AsymmetricMatchersContaining} from 'vitest'
+import {
+  type Assertion,
+  type AsymmetricMatchersContaining,
+  type expect,
+} from 'vitest'
 import {type TestingLibraryMatchers} from './matchers'
 
 export {}
--- a/types/vitest.d.ts
+++ b/types/vitest.d.ts
@@ -1,11 +1,16 @@
-import {type expect} from 'vitest'
+/* eslint-disable @typescript-eslint/no-shadow */
+
+import {type Assertion, type AsymmetricMatchersContaining} from 'vitest'
 import {type TestingLibraryMatchers} from './matchers'
 
 export {}
-declare module '@vitest/expect' {
-  interface JestAssertion<T = any>
+
+// https://vitest.dev/guide/extending-matchers.html
+declare module 'vitest' {
+  interface Assertion<T = any>
     extends TestingLibraryMatchers<
       ReturnType<typeof expect.stringContaining>,
       T
     > {}
+  interface AsymmetricMatchersContaining extends TestingLibraryMatchers {}
 }


ecoms-ye:
Confirmed in my office computer (a pnpm monorepo with the [With Vitest](https://github.com/testing-library/jest-dom?tab=readme-ov-file#with-vitest) and [With TypeScript](https://github.com/testing-library/jest-dom?tab=readme-ov-file#with-typescript) setup): 

- Switch to the fork version by changing `package.json` to `"@testing-library/jest-dom": "https://github.com/nyaapass/jest-dom"`, the error can be fixed:
    - image
- When back to the original `@testing-library/jest-dom@6.4.2` version, the #515 error would be shown again: 
    - image

Dalcvi:
What is the reason that this change is not being merged in? Vitest removed `@vitest/expect` interface, so when `@testing-library/jest-dom/vitest` is trying to extend `@vitest/expect` interface, it just does nothing. By the way I see it, this change actually corrects the way you are expected to extend the interface since 0.x.x version, so what is the hold up here?

types/vitest.d.ts
```diff
@@ -1,11 +1,20 @@
-import {type expect} from 'vitest'
+/* eslint-disable @typescript-eslint/no-shadow */
+
+import {
+  type Assertion,
+  type AsymmetricMatchersContaining,
```
Dalcvi, line 5: These two are unused.

You actually type out the name but don't use it in the line 14 and 19.

Dalcvi (changes requested):

Dalcvi:
> What is the reason that this change is not being merged in? Vitest removed `@vitest/expect` interface, so when `@testing-library/jest-dom/vitest` is trying to extend `@vitest/expect` interface, it just does nothing. By the way I see it, this change actually corrects the way you are expected to extend the interface since 0.x.x version, so what is the hold up here?

Alright, other PR was changed with these changes https://github.com/testing-library/jest-dom/commit/06d19ba759a5ff4b51417cc9a7ecb72c785eb74d waiting for release I guess

gnapse:
I guess we can close this as superseded by #612

Issue #481, opened 2022-09-22
Support buttons with checkbox role for 'toBeRequired'


CarlosHdez:
### Describe the feature you'd like:

Hi,
I'm using Radix checkboxes https://www.radix-ui.com/docs/primitives/components/checkbox#root, it renders a button with `role="checkbox"` and supports required by adding `aria-required="true"` but I can't test it because in here https://github.com/testing-library/jest-dom/blob/main/src/to-be-required.js#L6 does not include either button as a valid form tag or checkbox as a valid role.
Not sure if this would be a bug or just a feature requirement

thewanionly:
I've run into this issue as well using shadcn which is essentially using radix components. Any updates on this one? It would be good to support this. 

For now, as a workaround, I used `toHaveAttribute` matcher and disabled the lint rule:
```ts
 // eslint-disable-next-line jest-dom/prefer-required
expect(checkbox).toHaveAttribute('aria-required', 'true');
```

gnapse:
This is a valid issue.

It seems that jest-dom does not implement this part of what's [documented in MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-required):

> When form controls are created using non-semantic elements, such as a [<div>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/div) with a [role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles) of [checkbox](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role), the `aria-required` attribute should be included, with a value of `true`, to indicate to assistive technologies that user input is required on the element for the form to be submittable. The `aria-required` attribute can be used with HTML form elements; it is not limited to elements that have an ARIA role assigned.

JatinRanka:
Hey @gnapse, I would like to fix this issue. 

I just wanted to confirm whether adding a new function in [to-be-required](https://github.com/testing-library/jest-dom/blob/main/src/to-be-required.js) file that checks if the element has `aria-required` with the value true should do the job, right? Is there anything else that I need to take care of (apart from test)?

PS: This is going to be my first contribution to this repo. So, any help would be appreciated :)

gnapse:
> I just wanted to confirm whether adding a new function in [to-be-required](https://github.com/testing-library/jest-dom/blob/main/src/to-be-required.js) file that checks if the element has `aria-required` with the value true should do the job, right?

[...]

PR #590, opened 2024-03-12
fix: updates support for aria-required attribute in `to-be-required`

JatinRanka:
**What**:

The non-semantic HTML element fails `toBeRequired` method even though aria-required is set as `true` - #481 

It does not implement this part of what's [documented in MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-required):

> When form controls are created using non-semantic elements, such as a [\<div\>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/div) with a [role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles) of [checkbox](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role), the `aria-required` attribute should be included, with a value of `true`, to indicate to assistive technologies that user input is required on the element for the form to be submittable. The `aria-required` attribute can be used with HTML form elements; it is not limited to elements that have an ARIA role assigned.

**Why**:

The current code only only adheres to `aria-required` attribute for a set of HTML elements

**How**:

Updated the `toBeRequired` method to not depend on tag type or role type for elements with `aria-required` attribute

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -379,6 +379,7 @@ attribute.
 <textarea data-testid="textarea" required></textarea>
 <div data-testid="supported-role" role="tree" required></div>
 <div data-testid="supported-role-aria" role="tree" aria-required="true"></div>
+<div data-testid="without-aria-role" aria-required="true"></div>
 ```
 
 ```javascript
@@ -392,6 +393,7 @@ expect(getByTestId('select')).toBeRequired()
 expect(getByTestId('textarea')).toBeRequired()
 expect(getByTestId('supported-role')).not.toBeRequired()
 expect(getByTestId('supported-role-aria')).toBeRequired()
+expect(queryByTestId('without-aria-role')).toBeRequired()
 ```
 
 <hr />
--- a/README.md
+++ b/README.md
@@ -392,7 +392,6 @@ expect(getByTestId('select')).toBeRequired()
 expect(getByTestId('textarea')).toBeRequired()
 expect(getByTestId('supported-role')).not.toBeRequired()
 expect(getByTestId('supported-role-aria')).toBeRequired()
-expect(queryByTestId('without-aria-role')).toBeRequired()
 ```
 
 <hr />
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -379,6 +379,7 @@ attribute.
 <textarea data-testid="textarea" required></textarea>
 <div data-testid="supported-role" role="tree" required></div>
 <div data-testid="supported-role-aria" role="tree" aria-required="true"></div>
+<div data-testid="without-aria-role" aria-required="true"></div>
```
gnapse, line 382: I'm not sure that it should be supported in any element with `aria-required`. It should still validate that it is used on an element with an aria role, and particularly with a role that supports the `aria-required` attribute.

In particular, check the [Associated roles](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-required#associated_roles) section in the MDN documentation:

> Used in roles:
>
> - [checkbox](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role)
> - [combobox](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role)
> - [gridcell](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/gridcell_role)
> - [listbox](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role)
> - [radiogroup](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/radiogroup_role)
> - [spinbutton](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/spinbutton_role)
> - [textbox](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/textbox_role)
> - [tree](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tree_role)

So this new functionality should only acknowledge the attribute if the element on which it is applied has one of these roles.

gnapse (changes requested):

README.md
JatinRanka: @gnapse I have addressed the review changes.  Can you please review it again.

gnapse (approved):

Issue #592, opened 2024-03-15
Cost effective lodash


nike1v:
Please replace lodash with it's smaller representatives, there are only 4 imports, this shouldn't take long.
Appreciate it!

gnapse:
Contributions are welcome on this one. I'm unlikely to do it anytime soon.

Also, since jest-dom is not normally used in a production web app, but in the development environment only, the bundle size is of less concern. But I'm open to doing better.

Idea: we could cross-reference our use of lodash with https://youmightnotneed.com/lodash/

gnapse:
Closed via https://github.com/testing-library/jest-dom/pull/593

PR #593, opened 2024-03-22
refactor: remove lodash

re-taro:
**What**: 

Replace lodash logic and remove it from deps

**Why**: 

To resolve this issue -> #592 

**How**:

https://youmightnotneed.com/lodash/

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/to-have-form-values.js
+++ b/src/to-have-form-values.js
@@ -1,16 +1,10 @@
-import isEqualWith from 'lodash/isEqualWith.js'
-import uniq from 'lodash/uniq.js'
 import escape from 'css.escape'
-import {
-  checkHtmlElement,
-  compareArraysAsSet,
-  getSingleElementValue,
-} from './utils'
+import {checkHtmlElement, getSingleElementValue} from './utils'
 
 // Returns the combined value of several elements that have the same name
 // e.g. radio buttons or groups of checkboxes
 function getMultiElementValue(elements) {
-  const types = uniq(elements.map(element => element.type))
+  const types = [...new Set(elements.map(element => element.type))]
   if (types.length !== 1) {
     throw new Error(
       'Multiple form elements with the same name must be of the same type',
@@ -69,9 +63,15 @@ export function toHaveFormValues(formElement, expectedValues) {
   }
   const formValues = getAllFormValues(formElement)
   return {
-    pass: Object.entries(expectedValues).every(([name, expectedValue]) =>
-      isEqualWith(formValues[name], expectedValue, compareArraysAsSet),
-    ),
+    pass: Object.entries(expectedValues).every(([name, expectedValue]) => {
+      if (Array.isArray(formValues[name]) && Array.isArray(expectedValue)) {
+        return [...new Set(formValues[name])].every(v =>
+          new Set(expectedValue).has(v),
+        )
+      } else {
+        return formValues[name] === expectedValue
+      }
+    }),
     message: () => {
       const to = this.isNot ? 'not to' : 'to'
       const matcher = `${this.isNot ? '.not' : ''}.toHaveFormValues`
--- a/src/to-have-form-values.js
+++ b/src/to-have-form-values.js
@@ -1,8 +1,9 @@
+import isEqualWith from 'lodash/isEqualWith'
 import escape from 'css.escape'
 import {
   checkHtmlElement,
   getSingleElementValue,
-  isEqualWithArraysAsSets,
+  compareArraysAsSet,
 } from './utils'
 
 // Returns the combined value of several elements that have the same name
@@ -68,7 +69,7 @@ export function toHaveFormValues(formElement, expectedValues) {
   const formValues = getAllFormValues(formElement)
   return {
     pass: Object.entries(expectedValues).every(([name, expectedValue]) =>
-      isEqualWithArraysAsSets(formValues[name], expectedValue),
+      isEqualWith(formValues[name], expectedValue, compareArraysAsSet),
     ),
     message: () => {
       const to = this.isNot ? 'not to' : 'to'
--- a/src/to-have-form-values.js
+++ b/src/to-have-form-values.js
@@ -1,5 +1,9 @@
 import escape from 'css.escape'
-import {checkHtmlElement, getSingleElementValue} from './utils'
+import {
+  checkHtmlElement,
+  getSingleElementValue,
+  isEqualWithArraysAsSets,
+} from './utils'
 
 // Returns the combined value of several elements that have the same name
 // e.g. radio buttons or groups of checkboxes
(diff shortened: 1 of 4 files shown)


src/to-have-form-values.js
```diff
@@ -69,9 +63,15 @@ export function toHaveFormValues(formElement, expectedValues) {
...
-    ),
+    pass: Object.entries(expectedValues).every(([name, expectedValue]) => {
+      if (Array.isArray(formValues[name]) && Array.isArray(expectedValue)) {
+        return [...new Set(formValues[name])].every(v =>
+          new Set(expectedValue).has(v),
+        )
+      } else {
+        return formValues[name] === expectedValue
+      }
+    }),
```
gnapse, line 74: The one thing lost here is readability. It is easier to read the previous version.

What I suggest is that you create a `function isEqualWith` (or `isEqualWithArraysAsSets`) in the `utils.js` module, that would encapsulate this functionality. Makes sense?

src/to-have-value.js
```diff
@@ -30,7 +24,9 @@ export function toHaveValue(htmlElement, expectedValue) {
 
   return {
     pass: expectsValue
-      ? isEqualWith(receivedValue, expectedValue, compareArraysAsSet)
+      ? Array.isArray(receivedValue) && Array.isArray(expectedValue)
+        ? [...new Set(receivedValue)].every(v => new Set(expectedValue).has(v))
+        : receivedValue === expectedValue
```
gnapse, line 29: And reuse the `isEqualWith` function here.

gnapse (approved):
Looks good. Though, I'm suggesting an improvement to not lose readability of the code.

gnapse (changes requested):
I meant to request changes instead, due to my suggestion. Sorry.

re-taro:
Good suggestion. I will revise it!

src/utils.js
```diff
@@ -220,6 +220,14 @@ function toSentence(
   )
 }
 
+function isEqualWithArraysAsSets(arr1, arr2) {
+  if (Array.isArray(arr1) && Array.isArray(arr2)) {
+    return [...new Set(arr1)].every(v => new Set(arr2).has(v))
+  } else {
+    return arr1 === arr2
+  }
+}
```
gnapse, line 229: I'm now having second thoughts about discarding the use of lodash's `isEqual` entirely. For a few reasons:

- It is a battle-tested implementation that covers edge cases that we might not be covering here (see below for one of these cases).
- It is not featured in the website https://youmightnotneed.com/lodash/, meaning that they do not have any recommended in-house implementation that replaces the lodash function.

## Edge case not covered here.

I think that the function in lodash is probably recursive. In the `else` part it probably calls itself up:

```suggestion
function isEqualWithArraysAsSets(arr1, arr2) {
  if (Array.isArray(arr1) && Array.isArray(arr2)) {
    return [...new Set(arr1)].every(v => new Set(arr2).has(v))
  } else {
    return isEqualWith(arr1, arr2)
  }
}
```

Also, probably the lodash implementation avoids a potential infinite recursion problem if an object has references to part of itself, like this:

```js
const obj = { list: [obj] }
```

But probably we do not need to cover that edge case here. So maybe doing the change suggested above is enough, and we can indeed ditch lodash entirely.

What do you think?

gnapse, line 229: > [!NOTE]
> _Before diving into my long comment below, a caveat: maybe our `isEqual` implementation does not need the features that I'm highlighting below that are missing compared to lodash's `isEqual`. Let me know._

I'm now having second thoughts about discarding the use of lodash's `isEqual` entirely. For a few reasons:

- It is a battle-tested implementation that covers edge cases that we might not be covering here (see below for one of these cases).
- It is not featured in the website https://youmightnotneed.com/lodash/, meaning that they do not have any recommended in-house implementation that replaces the lodash function.

## Edge cases not covered here

I think that the function in lodash is also able to compare two objects for equality (i.e. looping over the keys and comparing that equal keys have equal values). Furthermore, that last part about comparing if equal keys have equal values is probably recursive: that is, it calls itself with the values of the two keys:

I asked ChatGPT and this is what it came up with (preserving our array-as-sets comparison as well):

```js
function isEqual(obj1, obj2) {
  // Check if both objects are arrays
  if (Array.isArray(obj1) && Array.isArray(obj2)) {
    // Sort the arrays before comparison
    const arraySet1 = [...new Set(arr1)]
    const arraySet2 = [...new Set(arr2)]

    // Compare the sorted arrays
    return arraySet1.every(v => arraySet2.has(v))
  }

  // Check if both arguments are objects
  if (typeof obj1 === 'object' && typeof obj2 === 'object') {
    // Get the keys of the objects
    const keys1 = Object.keys(obj1);
    const keys2 = Object.keys(obj2);

    // Check if the number of keys is the same
    if (keys1.length !== keys2.length) {
      return false;
    }

    // Check if all keys and their corresponding values are equal
    for (let key of keys1) {
      // Recursive call to isEqual for nested objects
      if (!isEqual(obj1[key], obj2[key])) {
        return false;
      }
    }

    // If all keys and values are equal, return true
    return true;
  }
  
  // If the arguments are not objects, compare them directly
  return obj1 === obj2;
}
```

And even then, probably we'd need to come up with a way to compare the array elements via `isEqual` recursively, which this implementation is not doing.

re-taro: I have one suggestion. However, this proposal is not a complete solution to the original issue. 
- Use lodash-es
- Re-implement the isEqual equivalent of lodash
  - It is not maintainable.

gnapse: I agree we can keep `isEqual`. If it's via `lodash-es` that's great. The goal should not blindly be about removing lodash entirely. Specially given that jest-dom is not normally a dependency that gets into a web app's bundle, given that it is mostly a dev dependency only.

I'd appreciate if you can do that change and we can merge this. We can leave trace in this PR and the original issue, that it was closed without fully removing lodash, but minimizing its usage to where is really needed.

Thanks!

re-taro: Sorry, this is just a simple question I have, but in the original code, where isEqualWith was used, what could be in the values being compared? Because the documentation for lodash.isEqual says `This method supports comparing arrays, array buffers, booleans, date objects, error objects, maps, numbers, Object objects, regexes, sets, strings, symbols, and typed arrays.` I personally feel that this is overkill, although I do not have a full understanding of the code or design.
I would appreciate it if you could let me know so that I can provide a better code.

gnapse: I know that the lodash version is surely covering use cases that we do not need, but on the flip side, a custom-made alternative may miss cases that we do need to cover. I'm not 100% sure that our test cases are iron-clad in covering all potential use cases. Hence, I'd rather keep `isEqual`, even if it means not totally ditching lodash.

I do not think that the goal should be to blindly avoid lodash. Specially in this library that's not going to be part of a web app bundle. So I'd rather stay on the safe side, and keep it.

Makes sense?

re-taro: I see, I'm certainly in favor of taking them down safely for the sake of trust.

gnapse: Thanks!

gnapse (approved):

clyncha:
One of my codebases is now having an issue with lodash/isEqualWith.js since dependabot bumped @testing-library/jest-dom from 6.4.2 to 6.4.3.

It looks like there was a relevant change in this PR

Screenshot 2024-05-03 at 4 38 30 PM

PR #602, opened 2024-06-05
Pr/starting style

jantimon:
**What**:

Upgraded the `@adobe/css-tools` dependency to include the fix for parsing the [`@starting-style` CSS feature](https://developer.mozilla.org/en-US/docs/Web/CSS/@starting-style), which was causing a `missing '}'` error.

**Why**:

The current version of `@adobe/css-tools` used by `testing-library/jest-dom` fails to parse the `@starting-style` CSS feature, resulting in a parsing error. The upgrade includes a fix for this issue, allowing for the correct parsing of CSS files using `@starting-style`.

**How**:

The changes were implemented by updating the `package.json` file to use the latest version of `@adobe/css-tools` which includes the fix from the PR I submitted to `@adobe/css-tools` (https://github.com/adobe/css-tools/pull/323).

**Checklist**:

- [ ] Documentation
- [ ] Tests (N/A due to dependency on `jsdom` and `CSSOM` updates)
- [ ] Updated Type Definitions (N/A)
- [x] Ready to be merged
      

**Additional Comments**:

Unfortunately, I could not add a test case for the `@starting-style` feature because `jsdom 20` uses CSSOM 0.6.0 for CSS parsing, which also has a similar issue. 
I fixed that already in this PR (https://github.com/rrweb-io/CSSOM/pull/1), but the latest version of `jsdom` cannot be merged into `jest`. See https://github.com/jestjs/jest/pull/14846#issuecomment-1885785901 for details

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
   "author": "Ernesto Garcia <gnapse@gmail.com> (http://gnapse.github.io)",
   "license": "MIT",
   "dependencies": {
-    "@adobe/css-tools": "^4.3.2",
+    "@adobe/css-tools": "^4.4.0",
     "@babel/runtime": "^7.9.2",
     "aria-query": "^5.0.0",
     "chalk": "^3.0.0",
--- a/src/__tests__/to-have-style.js
+++ b/src/__tests__/to-have-style.js
@@ -251,33 +251,4 @@ describe('.toHaveStyle', () => {
       })
     })
   })
-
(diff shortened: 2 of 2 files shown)


gnapse (approved):

gnapse:
@all-contributors, please add @jantimon for bug

gnapse:
@all-contributors please add @jantimon for bug

PR #612, opened 2024-07-15
adjust vitest ambient declaration file according to vitest docs

vorant94:
**What**:

Changed the way of patching `vitest` types with `jest-dom` custom matchers

**Why**:

The current approach uses `@vitest/expect` module, which is considered to be internal implementation details by the `vitest` team according to comments [here](https://github.com/vitest-dev/vitest/pull/6121#discussion_r1677098589), so it shouldn't be used to accomplish the goal

**How**:

New approach follows the latest official `vitest` documentation. There was also docs fix that I proposed and got approved recently, so the `vitest` docs website itself isn't reflecting the latest state. To verify that new approach is aligned with docs, you can check this [PR](https://github.com/vitest-dev/vitest/pull/6121)

Along the way I removed `export {}` statement since it isn't needed when there are `import` statements are present in the same file

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Ready to be merged


gnapse (approved):

gnapse:
@all-contributors please add @vorant94 for code

Issue #620, opened 2024-08-02
Property 'toHaveTextContent' does not exist on type 'Assertion<HTMLElement>'.


drichar:
- `@testing-library/jest-dom` version: 6.4.8
- `node` version: 20.14.0
- `vitest` version: 2.0.4
- `pnpm` version: 9.4.0
- `@testing-library/react` version: 16.0.0

### Relevant code or config:

```js
describe('ellipseAddressJsx', () => {
  it('should return ellipsed address with specified width', () => {
    const address = 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'
    const width = 4

    const { container } = render(ellipseAddressJsx(address, width))

    expect(container).toHaveTextContent('aaaa...aaaa')
  })
})
```

### What you did:

Upgraded from `vitest` 2.0.3 from 2.0.4

### What happened:

After upgrading to `vitest@2.0.4` this TypeScript error appeared in every `@testing-library/jest-dom` assertion:
<img width="715" alt="image" src="https://github.com/user-attachments/assets/642f99f6-2d8b-4f32-81f3-b52b153c88aa">

### Reproduction:

https://codesandbox.io/p/devbox/silly-panna-vx86jv

See src/utils.test.tsx

### Problem description:

Following the instructions for using `@testing-library/jest-dom` with Vitest provided [here](https://github.com/testing-library/jest-dom#with-vitest) no longer seems to work in the latest version. (Confirmed this is also the case for `vitest@2.0.5`)

### Suggested solution:

As suggested here: https://github.com/testing-library/jest-dom/issues/546#issuecomment-1800436478

In tsconfig.json:
```json
{
  "compilerOptions": {
    "types": ["@testing-library/jest-dom"]
  },
}
```

This resolves the issue.

gnapse:
Does https://github.com/testing-library/jest-dom/pull/621 fixes the problem? If I understand correctly, this is something that consumers of the library would need to do in their own repos, so all we can do is document this in the README, right?

andykenward:
Instead of changing your `"types"`, add your `vitest.setup.ts` file to the `"include"` array in your `tsconfig.json`.

tsconfig.json
```json
{
  ...
  "include": ["src", "vitest.setup.ts"],
}
```

[...]

PR #621, opened 2024-08-02
docs: Update tsconfig setup instructions in the README

gnapse:
**What**:

Fixes https://github.com/testing-library/jest-dom/issues/620

**Checklist**:

- [x] Documentation
- [x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,15 +170,20 @@ setupFiles: ['./vitest-setup.js']
 If you're using TypeScript, make sure your setup file is a `.ts` and not a `.js`
 to include the necessary types.
 
-You will also need to include your setup file in your `tsconfig.json` if you
-haven't already:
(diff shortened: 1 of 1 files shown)


README.md
andykenward, line 1: Instead of these changes, we should add `"./vitest.setup.ts"` next to `"./jest-setup.ts"`, to make it a bit more obvious.

It appears that a few users are missing out the setup step for TypeScript when using a `*.setup.ts` file. Perhaps we should add a reference to it in other sections and also make the below line bold?

>If you're using TypeScript, make sure your setup file is a .ts and not a .js to include the necessary types

>**If you're using TypeScript, make sure your setup file is a .ts and not a .js to include the necessary types**

andykenward (changes requested):

PR #636, opened 2024-10-07
fix: remove recursive type reference in vitest types

### LINKED ISSUE #629: Vitest typings are broken in 6.5.0
- `@testing-library/jest-dom` version: 6.5.0
- `node` version: 20.15.1
- `jest` (or `vitest`) version: Vitest 2.0.5
- `npm` (or `yarn`) version: npm 10.8.2

### Relevant code or config:

See reproduction repo below.

### What you did:

Followed the setup instructions for Vitest in the readme.

### What happened:

Running the TS checker in a minimal environment produces the following errors:

```
node_modules/@testing-library/jest-dom/types/vitest.d.ts:7:32 - error TS2339: Property 'stringContaining' does not exist on type 'ExpectStatic'.

7       ReturnType<typeof expect.stringContaining>,
                                 ~~~~~~~~~~~~~~~~

node_modules/@testing-library/jest-dom/types/vitest.d.ts:10:13 - error TS2310: Type 'AsymmetricMatchersContaining' recursively references itself as a base type.

10   interface AsymmetricMatchersContaining
               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/@testing-library/jest-dom/types/vitest.d.ts:12:32 - error TS2339: Property 'stringContaining' does not exist on type 'ExpectStatic'.

12       ReturnType<typeof expect.stringContaining>,
                                  ~~~~~~~~~~~~~~~~
```

### Reproduction:

I've isolated the problem to a barebones setup:

https://github.com/billyjanitsch/vitest-jest-dom-types

### Problem description:

https://github.com/testing-library/jest-dom/pull/612 (which was released in 6.5.0) updated the typings for Vitest in a way that seems to be incompatible with Vitest's own definitions. I understand that that PR was an attempt to align the typings with the official recommendations, but it seems to be broken.

### Suggested solution:

I'd suggest reverting https://github.com/testing-library/jest-dom/pull/612 until it can be tested and fixed.

For a long-term fix, this is just a guess based on the TS error and looking at the type defs, but I think the problem might be related to importing `type expect` from `vitest` and then using that type when extending the `vitest` module type. This recursive self-reference might not be allowed/supported by the TS compiler.

cc @vorant94

G-Rath:
**What**:

Replaces the dynamic lookup of the return type of `expect.stringContaining` with the underlying type

**Why**:

This resolves the TypeScript errors introduced by #612, as described in #629

**How**:

By replacing the use of `expect.stringContaining` with a concrete type - I don't actually know why `ReturnType` was being used in the first place given that `expect.stringContaining` has a return type of `any` and I don't really see that changing anytime in the future.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged
      

Resolves #632
Resolves #629

### DIFF
--- a/types/vitest.d.ts
+++ b/types/vitest.d.ts
@@ -1,15 +1,15 @@
-import {type expect} from 'vitest'
+
 import {type TestingLibraryMatchers} from './matchers'
 
 declare module 'vitest' {
   interface Assertion<T = any>
     extends TestingLibraryMatchers<
-      ReturnType<typeof expect.stringContaining>,
+      any,
       T
     > {}
   interface AsymmetricMatchersContaining
     extends TestingLibraryMatchers<
-      ReturnType<typeof expect.stringContaining>,
+      any,
       any
     > {}
 }


G-Rath:
@gnapse would it be possible to have this reviewed? as currently all typescript types are broken meaning we've having to patch this manually for our CIs to pass 😅

gnapse (approved):

gnapse:
@all-contributors please add @G-Rath for code

gnapse:
@G-Rath can you look into it? I have no access to an environment with jest-dom and vitest right now.

greysteil:
FWIW I'm not seeing the errors @BR0kEN- is - the change is working well for me (and I had the issue @G-Rath had with `jest-dom` and `vitest` before).

BR0kEN-:
Thanks for the fast reply and apologies, I had to not post that before a little of research.

I had no `@testing-library/jest-dom` added to the `compilerOptions.types` of the `tsconfig.json` and it worked. After this change, which is logical, it stopped. Fortunately, the fix is easy.

Issue #645, opened 2024-10-18
Vitest typings for jest-dom are broken in 6.6.2 when using "globalMode"


InfiniteXyy:
- `@testing-library/jest-dom` version: 6.6.2
- `node` version: 20.10.0
-  vitest version: 2.1.3
- `npm` (or `yarn`) version: 10.2.3

### Relevant code or config / Reproduction:

minimun reproduce demo: https://stackblitz.com/edit/vitejs-vite-ydhnrx?file=src%2FApp.test.tsx

### What you did:

Bootstramp a starter repo with vitest & testing-library (vitest should be in global mode, and there is no `import xxx from 'vitest'` in the whole project)

### Problem description / What happened:

The `toHaveTextContent` should be a matcher of expect, and also is working in runtime, but a type error is thrown.

You can run `npm run build` to see the error

### Suggested solution:

According to the documents of [vitest](https://vitest.dev/guide/extending-matchers.html), to extend a matcher, we have to import the `"vitest"` module otherwise the interface merge might not work as expected
So we could add this line of code to top of the definition

```ts
import "vitest"
```

You can try adding this in the stackbliz repo's dependency (jumping to the source code by ctrl + click the import) and run `npm run build` again, the type error should be fixed

mrmckeb:
This was likely introduced by the change in this commit:
https://github.com/testing-library/jest-dom/pull/636

LucidityDesign:
Adding `"types": ["@testing-library/jest-dom"]` to tsconfig solves this issue for me.

InfiniteXyy:
> Adding `"types": ["@testing-library/jest-dom"]` to tsconfig solves this issue for me.

Yes, I think it's working, but in the source code of `"@testing-library/jest-dom"` it's referencing jest, which is not so correct.
```ts
/// <reference types="jest" />
```

If you happen to have `@types/jest` installed in project, it will make the expect become `jest.Expect`, for example in a monorepo

ChaseMarcum:
I am running into the same issue with `toHaveNoViolations` matcher for vitest-axe.

- `@testing-library/jest-dom` version: 6.6.2
- `node` version: 20.17.0
- `vitest` version: 2.0.0
- `vitest-axe` version: 0.1.0
- `npm` version: 10.8.2

### Relevant code or config:

```ts
// vitest-setup.js
import * as matchers from "vitest-axe/matchers";
import { expect } from "vitest";
expect.extend(matchers);
```

```ts
[...]

PR #646, opened 2024-10-18
fix: add vitest import when extending vitest matchers

InfiniteXyy:
**What**:
https://github.com/testing-library/jest-dom/issues/645
Fix a vitest typing issue when using global mode

**Why**:
If there is no import of "vitest" in the whole project (for example, vitest is in globalMode), the type merging might not work.

**How**:
By adding an import of "vitest" when extending the matcher type

reference: https://vitest.dev/guide/extending-matchers.html

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/types/vitest.d.ts
+++ b/types/vitest.d.ts
@@ -1,4 +1,4 @@
-
+import 'vitest'
 import {type TestingLibraryMatchers} from './matchers'
 
 declare module 'vitest' {


G-Rath (approved):
sorry about this! I think the fix here is right

cc @gnapse

gnapse (approved):

gnapse:
@all-contributors please add @InfiniteXyy for code, bug

dnafication:
This is causing issues like #662. Is there a workaround for this?

dnafication:
The issue occurs on triggering tests using vitest. Just running npx vitest
results in the error

dnafication:
Turns out the issue was due to a leftover import in my vitest.setup.ts for cleaning up after each test. I still see it in the [setup instruction for svelte](https://testing-library.com/docs/svelte-testing-library/setup#:~:text=Add%20a%20vitest%2Dsetup.js%20file%20to%20optionally%20set%20up%20%40testing%2Dlibrary/jest%2Ddom%20expect%20matchers).

```js
import "@testing-library/jest-dom/vitest"
```

PR #671, opened 2025-06-06
fix: replace lodash with es-toolkit

manudeli:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged


okinawaa:
@gnapse Hello. 
I'm sorry while you're busy.
Can you check this PR? The context is well written in the description.

wojtekmaj (approved):
LGTM, and it looks like choosing compat function is the right choice here. From what I see, every isEqualWith call is being made with a third argument being compareArraysAsSet util, which might return undefined when we're not comparing two arrays, which would trigger the compat behavior that comes after this line evaluates to false: https://github.com/toss/es-toolkit/blob/main/src/compat/predicate/isEqualWith.ts#L48

PR #676, opened 2025-08-05
refactor: replace `lodash` with `es-toolkit`

MichaelDeBoey:
`lodash` is unmaintained, big and hard to tree-shake.
`es-toolkit` is a drop-in replacement

---

I'm sure people from the wider @e18e ecosystem cleanup (like @43081j, @benmccann & @outslept) will be very happy to see these kind of changes as well


beeequeue:
I would recommend moving also this dependency to the devDeps and bundling it by marking it as not external in the rollup config
This would reduce the code that people have to download from 2.5mb to 3kb

timdeschryver (approved):

PR #678, opened 2025-08-05
chore: add tests for Node 22 & 24

MichaelDeBoey:


MatanBobi:
Thanks @MichaelDeBoey!
I think we can also safely remove node 14 and node 16 (for tests at least). Thoughts?

MichaelDeBoey:
@MatanBobi I don't think we should remove those if we're not actually dropping support for them

PR #731, opened 2026-07-10
feat: add toContainAnyBy* and toContainOneBy* query matchers

### LINKED ISSUE #548: Support assertions for each DOM query
## Describe the feature you'd like:

TL;DR:
```ts
expect(element).toContainOneByRole("heading", {
    name: "Attention"
});
```

Please consider adding jest assertions for each of the eight query types provided by `@testing-library/dom`. Assertions based directly on DOM queries would make [this advice](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-get-variants-as-assertions) much easier to follow.

> Advice: If you want to assert that something exists, make that assertion explicit.

### Problem Statement

In many tests, asserting on both the existence _and absence_ of elements is key to proving correct behavior. While it's true that `getBy*` and friends already throw excellent diagnostic information when something can't be found:

* Relying on the throwing side effect for the purpose of assertion does not make the intent of a test clear.
* The errors these queries throw don't help when asserting on the _absence_ of an element.
* Wrapping queries in fake assertions to make the intent clear results in spurious test feedback.

### Fabricated Example

An app needs to show and hide a warning within a dialog as validity changes. This is what the test might look like today.

```ts
const dialog = screen.getByRole("dialog");
const input = within(dialog).getByLabelText("Foo");
userEvent.type(input, "INVALID");

// Don't forget to use `getBy*` here, or we'll see a confusing null error.
expect(within(dialog).getByRole("heading", {name: "Attention"})).toBeInTheDocument();

userEvent.type(input, "VALID");

// Don't forget to use `queryBy*` here or it will fail before it can check for absence.
expect(within(dialog).queryByRole("heading", {name: "Attention"})).not.toBeInTheDocument();
```

### Drawbacks

Having assertions that are tightly integrated with DOM queries would require a dependency on `@testing-library/dom` which does not exist today. That said, the library is called `@testing-library/jest-dom`, which implies the two are connected.

### Proposal

Using "by role" as an example:

- Add `toContainOneByRole` which passes when there is precisely one element. The negated version `not.toContainOneByRole` passes when there are zero or more than one elements. This has some similarities with `getByRole`.
- Add `toContainAnyByRole` which passes when there are one or more elements. The negated version `not.toContainAnyByRole` passes when there are zero elements. This has some similarities with `getAllByRole`.

The assertion arguments would have exact symmetry with each query function. For example:
```ts
toContainOneByRole(role: ByRoleMatcher, options?: ByRoleOptions): void;
toContainAnyByRole(role: ByRoleMatcher, options?: ByRoleOptions): void;
```

Repeat for the other seven query types in `@testing-library/dom`.

### Sample Implementation

A hastily put together implementation can be found in this gist: https://gist.github.com/benquarmby/7510252ab701669c2eaf3c0156dd680c. There are holes, but it works for demonstration purposes.

## Describe alternatives you've considered:

- Making a separate library, either official as `@testing-library/jest-dom-query` or unofficial.
- Copy and pasting the custom matchers around.

## Teachability, Documentation, Adoption, Migration Strategy:

Using the previous fabricated example, usage could look like this:

```ts
const dialog = screen.getByRole("dialog");
const input = within(dialog).getByLabelText("Foo");
userEvent.type(input, "INVALID");

expect(dialog).toContainOneByRole("heading", {name: "Attention"});

userEvent.type(input, "VALID");

expect(dialog).not.toContainAnyByRole("heading", {name: "Attention"});
```

benquarmby:
**What**:

This PR adds 16 new query matchers in the form `toContainAnyBy*` and `toContainOneBy*` covering all eight core Testing Library query types: `AltText`, `DisplayValue`, `LabelText`, `PlaceholderText`, `Role`, `TestId`, `Text`, and `Title`.

**Why**:

Resolves #548 which has the full context. TL;DR: Help consumers with:

> Advice: If you want to assert that something exists, make that assertion explicit.

**How**:

Because all 16 matchers are nearly identical, a single new `query-matchers.js` module uses a factory to build the matchers and then export them as an object. The `matchers.js` module then destructures them as named exports.

Even though the new module is only a hundred lines, this is a large PR due to the tests, type declarations and documentation. I will add more context via inline PR comments to help with the review.

> [!IMPORTANT]
> As discussed in https://github.com/testing-library/jest-dom/issues/548#issuecomment-1811319721, this implementation requires a peer dependency on `@testing-library/dom`, which is a breaking change. We must also drop Node < 18. The commit message has a BREAKING CHANGE footer to match.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged


src/query-matchers.js
```diff
@@ -0,0 +1,100 @@
...
+            elements.length,
+          ),
+        ]
+        const prettyElements = elements
+          .map(element => prettyDOM(element))
+          .filter(isString)
+
+        if (prettyElements.length) {
+          parts.push('Here are the matching elements:', ...prettyElements)
+        }
```
benquarmby, line 81: This is a similar behavior to [dom-testing-library/query-helpers.ts](https://github.com/testing-library/dom-testing-library/blob/e395d5b8ecd4ec95cc02540ddcf16571ecd33e0f/src/query-helpers.ts#L74-L81). The intent is to make these matchers as helpful diagnostically as `@testing-library/dom` itself.

src/query-matchers.js
```diff
@@ -0,0 +1,100 @@
...
+      },
+      pass,
+    }
+  }
+
+  // The function's .name must align with the matcher for the checkHtmlElement
+  // hint to render correctly.
+  Object.defineProperty(queryMatcher, 'name', {
+    value: matcherName,
+  })
```
benquarmby, line 93: Slightly quirky requirement here, so I added a code comment to explain. This is the existing code that our generated matcher function needs to work with: https://github.com/testing-library/jest-dom/blob/55c07ce5f1c489b5b9dc31a770a84d83a1178072/src/utils.js#L24-L28

package.json
```diff
@@ -103,6 +104,9 @@
     "typescript": "^5.1.6",
     "vitest": "^0.34.1"
   },
+  "peerDependencies": {
+    "@testing-library/dom": ">=8 <11"
+  },
```
benquarmby, line 109: This is not open ended because we can't predict what `@testing-library/dom` v11 and higher will ship in terms of breaking changes. Whenever that day comes, increasing this range to support more versions is a straight-forward, non-breaking feature.

types/matchers.d.ts
```diff
@@ -1,6 +1,26 @@
...
+  type DomQueryArgs<F extends DomQueryFunction = DomQueryFunction> = F extends (
+    container: HTMLElement,
+    ...args: infer A
+  ) => HTMLElement[]
+    ? A
+    : never
+
+  type DomQueryByBoundAttributeArgs = DomQueryArgs<AllByBoundAttribute>
+  type DomQueryByRoleArgs = DomQueryArgs<AllByRole>
+  type DomQueryByTextArgs = DomQueryArgs<AllByText>
```
benquarmby, line 22: The `DomQueryArgs<F>` utility type infers the argument types directly from `@testing-library/dom`. This ensures the matcher signatures stay in sync with the upstream queries automatically rather than duplicating them by hand.

types/matchers.d.ts
```diff
@@ -761,7 +781,7 @@ declare namespace matchers {
      * [testing-library/jest-dom#tohaveselection](https://github.com/testing-library/jest-dom#tohaveselection)
      */
     toHaveSelection(selection?: string): R
-    /*
+    /**
      * @description
```
benquarmby, line 785: Find a bug, fix a bug. I noticed these weren't being picked up as JSDoc blocks.

package.json
```diff
@@ -118,7 +122,11 @@
           "src/__tests__/*.js"
         ],
         "rules": {
-          "max-lines-per-function": "off"
+          "max-lines-per-function": "off",
+          "testing-library/no-container": "off",
+          "testing-library/no-render-in-setup": "off",
+          "testing-library/prefer-screen-queries": "off",
+          "testing-library/render-result-naming-convention": "off"
```
benquarmby, line 129: Needing to change these ESLint rules was a surprise, and definitely worth explaining.

This repo inherits its ESLint config from `eslint-config-kentcdodds/jest` via `kcd-scripts`. These testing-library rules are included, but they lie dormant until `@testing-library/dom` is a dependency. Well, guess who just added it?! 😅 

There are 500+ violations in this repo alone. Fixing those is well outside the scope of this PR, so I opted to disable them instead. That's the same state as before.

benquarmby (commented):
I added some inline comments to provide more context.

benquarmby:
A friendly ping on this one @gnapse, @kentcdodds. This PR follows through on the approach signed off on in #548. Sixteen new matchers covering the eight core queries (via factory) with full test coverage, type declarations, documentation, and a BREAKING CHANGE notation as discussed.

The concept sat for a while before we got here. It's been baking on internal test suites in the meantime, so confidence is high.

The CI workflows need a maintainer approval to trigger. Can you please unblock those? Then we can see the full results and the PR will be ready for review.

benquarmby:
Thanks for running the build. We need to drop old Node versions much like:
- https://github.com/testing-library/dom-testing-library/pull/1255.

I have made that change and included it in the breaking change footer:
> BREAKING CHANGE: @testing-library/dom is now a required peer dependency.
> The minimum supported Node.js version is now 22.

This also limits the peer range of `@testing-library/dom` to `>=10 <11`.

Can you please re-run validation when you get a chance?

kentcdodds:
Personally I'd drop Node 18 and 20 as well. They're no longer supported versions of Node anyway.

benquarmby:
> Personally I'd drop Node 18 and 20 as well. They're no longer supported versions of Node anyway.

Agreed and done. Breaking change footer adjusted. I'll leave it to you to adjust the required checks for this repo. Thanks again.

kentcdodds (approved):
Looks good. Thanks!

kentcdodds:
I messed up the release a bit, but I got this published as v7 and deprecated the mistaken release.

PR #733, opened 2026-08-08
fix: declare vitest as an optional peer dependency

### LINKED ISSUE #662: Cannot find package 'vitest'
- `@testing-library/jest-dom` version: 6.6.3
- `node` version: v22.13.1
- `jest` (or `vitest`) version: 3.0.5
- `npm` (or `yarn`) version: pnpm@9.15.1

### Relevant code or config:

```
; .npmrc
hoist=false
```

### What you did:

Tried running tests.

### What happened:

```
Error: Cannot find package 'vitest' imported from /path/to/project/ui/node_modules/.pnpm/@testing-library+jest-dom@6.6.3/node_modules/@testing-library/jest-dom/dist/vitest.mjs
```

### Suggested solution:

`vitest` is declared in the `devDependencies` of `@testing-library/jest-dom` but it must be declared in the `dependencies` or `peerDependencies` because `@testing-library/jest-dom/dist/vitest.mjs` is importing `vitest`s `expect` at test run time.

big-mon:
**What**:

Declare `vitest` as an optional peer dependency.

Fixes #662.

**Why**:

The `@testing-library/jest-dom/vitest` entry point imports `vitest` at runtime, but `vitest` is currently only a development dependency. Strict dependency layouts therefore cannot resolve the import from `jest-dom`, even when the consuming project directly depends on both packages.

This is reproducible with pnpm 11's global virtual store using `@testing-library/jest-dom@6.9.1` and `vitest@4.1.9`:

```text
Error: Cannot find package 'vitest' imported from .../@testing-library/jest-dom/dist/vitest.mjs
```

Keeping the peer optional avoids requiring or installing Vitest for Jest/Bun consumers. The range restores the previously supported `>= 0.32` contract.

**How**:

- Add `vitest` to `peerDependencies` as `>= 0.32`.
- Mark the peer optional in `peerDependenciesMeta`.
- Add a regression test for the published package metadata.

A packed build was also installed into a fresh pnpm 11 project with `enableGlobalVirtualStore: true`; pnpm bound the consumer's `vitest@4.1.9` to `jest-dom`, and the Vitest setup import passed without `packageExtensions`.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/package-json.js
+++ b/src/__tests__/package-json.js
@@ -1,0 +1,8 @@
+const packageJson = require('../../package.json')
+
+test('declares vitest as an optional peer dependency', () => {
+  expect(packageJson.peerDependencies).toMatchObject({vitest: '>= 0.32'})
+  expect(packageJson.peerDependenciesMeta).toMatchObject({
+    vitest: {optional: true},
+  })
+})
--- a/src/__tests__/package-json.js
+++ b/src/__tests__/package-json.js
@@ -1,8 +1,0 @@
-const packageJson = require('../../package.json')
-
-test('declares vitest as an optional peer dependency', () => {
-  expect(packageJson.peerDependencies).toMatchObject({vitest: '>= 0.32'})
-  expect(packageJson.peerDependenciesMeta).toMatchObject({
-    vitest: {optional: true},
-  })
-})
--- a/package.json
+++ b/package.json
@@ -105,7 +105,13 @@
     "vitest": "^0.34.1"
   },
   "peerDependencies": {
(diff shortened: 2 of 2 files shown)


src/__tests__/package-json.js
gnapse, line 1: I would say that having a test for this is not needed

gnapse (approved):
Thanks!

src/__tests__/package-json.js
big-mon: Removed it. Thanks!

PR #739, opened 2026-09-06
fix(to-have-focus): resolve the active element through shadow roots

### LINKED ISSUE #722: toHaveFocus doesn't work with shadow roots
- `@testing-library/jest-dom` version: 6.9.1                                                                                                                                                                                                                                            
  - `node` version: 24.15.0                                                                                                                                                                                                                                                               
  - `jest` version: 30.3.0                                                                                                                                                                                                                                                                
  - `npm` version: 11.12.1                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                          
  ### Relevant code or config:
                                                                                                                                                                                                                                                                                          
```js                                                          
  const host = document.createElement('div');
  const shadowRoot = host.attachShadow({ mode: 'open' });
  const input = document.createElement('input');
  shadowRoot.appendChild(input);                                                                                                                                                                                                                                                          
  document.body.appendChild(host);
                                                                                                                                                                                                                                                                                          
  input.focus();                                                  

  expect(input).toHaveFocus(); // fails
```

  ### What you did:

  Focused an `<input>` element inside a shadow root, then asserted `toHaveFocus()` on it.                                                                                                                                                                                                 
  
  ### What happened:                                                                                                                                                                                                                                                                      
                                                                  
```
  expect(element).toHaveFocus()

  Expected element with focus:
    <input />
  Received element with focus:
    <div />
```                                                                                                                                                                                                                                                                                    
  
  The `<div />` reported as focused is the shadow host, not the `<input>`.                                                                                                                                                                                                                
                                                                  
  ### Reproduction:

  https://github.com/julienw/jest-dom-to-have-focus-shadow-root

  Run `npm install && npm test`. Six tests pass and one fails — the failing test is the bug.                                                                                                                                                                                              
  
  ### Problem description:                                                                                                                                                                                                                                                                
                                                                  
  When focus moves into a shadow root, `document.activeElement` returns the **shadow host** rather than the element that actually has focus. `toHaveFocus` checks `element.ownerDocument.activeElement === element`, so it can never return `true` for any element inside a shadow root.  
  
  The actually-focused element is reachable via `shadowRoot.activeElement`, which the matcher does not consult.                                                                                                                                                                           
                                                                  
  ### Suggested solution:                                                                                                                                                                                                                                                                 
  
  Traverse shadow roots when resolving the active element:                                                                                                                                                                                                                                
                                                                  
  ```js
  function getDeepActiveElement(doc) {
    let active = doc.activeElement;
    while (active?.shadowRoot?.activeElement) {
      active = active.shadowRoot.activeElement;
    }
    return active;                                                                                                                                                                                                                                                                        
  }
  ```                                                                                                                                                                                                                                                                                    
                                                                  
  Replacing `element.ownerDocument.activeElement` with `getDeepActiveElement(element.ownerDocument)` in the [`toHaveFocus` implementation](https://github.com/testing-library/jest-dom/blob/main/src/to-have-focus.ts) would fix the issue for arbitrarily nested shadow roots.           


If the solution works for you I can do a PR with that.

FrancoKaddour:
Closes #722.

## Problem

`toHaveFocus` checks `element.ownerDocument.activeElement === element`. When focus moves into a shadow root, `document.activeElement` resolves to the shadow *host*, not the element that actually has focus (which lives at `shadowRoot.activeElement`). As a result, `toHaveFocus` can never pass for any element inside a shadow root:

```js
const host = document.createElement('div')
const shadowRoot = host.attachShadow({mode: 'open'})
const input = document.createElement('input')
shadowRoot.appendChild(input)
document.body.appendChild(host)

input.focus()
expect(input).toHaveFocus() // failed before this change
```

## Fix

Add a small `getActiveElement` helper that walks down nested shadow roots to find the truly focused element, and use it both in the `pass` check and in the failure message (so the "Received element with focus" diagnostic points at the real element rather than the host). For regular elements with no shadow root it degrades to `doc.activeElement`, so existing behavior is unchanged.

Thanks to @julienw for the clear report and the suggested approach.

## Tests

Added coverage for a focused element inside a shadow root and inside nested shadow roots. Both fail on `main` and pass with this change; the existing `.toHaveFocus` test is unaffected.

### DIFF
--- a/src/__tests__/to-have-focus.js
+++ b/src/__tests__/to-have-focus.js
@@ -21,3 +21,39 @@ test('.toHaveFocus', () => {
   expect(() => expect(focused).not.toHaveFocus()).toThrowError()
   expect(() => expect(notFocused).toHaveFocus()).toThrowError()
 })
+
+test('.toHaveFocus resolves the active element through shadow roots', () => {
+  const host = document.createElement('div')
+  document.body.appendChild(host)
+
+  const shadowRoot = host.attachShadow({mode: 'open'})
+  const focused = document.createElement('input')
+  const notFocused = document.createElement('input')
+  shadowRoot.appendChild(focused)
+  shadowRoot.appendChild(notFocused)
+
+  focused.focus()
+
+  expect(focused).toHaveFocus()
+  expect(notFocused).not.toHaveFocus()
+
+  expect(() => expect(focused).not.toHaveFocus()).toThrowError()
+  expect(() => expect(notFocused).toHaveFocus()).toThrowError()
+})
+
+test('.toHaveFocus resolves the active element through nested shadow roots', () => {
+  const outerHost = document.createElement('div')
+  document.body.appendChild(outerHost)
+
+  const outerShadow = outerHost.attachShadow({mode: 'open'})
+  const innerHost = document.createElement('div')
+  outerShadow.appendChild(innerHost)
+
+  const innerShadow = innerHost.attachShadow({mode: 'open'})
+  const focused = document.createElement('input')
+  innerShadow.appendChild(focused)
+
+  focused.focus()
+
+  expect(focused).toHaveFocus()
+})
--- a/src/to-have-focus.js
+++ b/src/to-have-focus.js
@@ -1,10 +1,25 @@
 import {checkHtmlElement} from './utils'
 
+// When focus moves into a shadow root, `document.activeElement` resolves to the
+// shadow host rather than the element that actually has focus, which lives at
+// `shadowRoot.activeElement`. Walk down nested shadow roots to find the truly
+// focused element. Falls back to `doc.activeElement` when no shadow root is
+// involved, so behavior is unchanged for regular elements.
+function getActiveElement(doc) {
+  let active = doc.activeElement
+
+  while (active?.shadowRoot?.activeElement) {
+    active = active.shadowRoot.activeElement
+  }
+
+  return active
+}
+
 export function toHaveFocus(element) {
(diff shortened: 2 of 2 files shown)


julienw:
Just thinking about that now: I've been using `. matches(":focus")` in my projects too. I wonder if that could be an easier and more elegant solution...

We could also add `.toHaveFocusWithin` using the css selector `:focus-within`.

FrancoKaddour:
Good thought — I tried `:focus` first too, but it doesn't hold up in jsdom, which is the environment most jest-dom users run in. Quick probe (input focused inside an open shadow root):

```
document.activeElement === input     → false   (the bug)
getActiveElement(document) === input → true    (this PR)
input.matches(':focus')              → false   ✗
```

jsdom doesn't propagate `:focus` across the shadow boundary, so `matches(':focus')` gives a false negative for exactly the case we're fixing. The explicit `shadowRoot.activeElement` walk sidesteps that by not relying on selector matching at all. (In a real browser `:focus` behaves better, but jest-dom has to work in jsdom.)

Re `toHaveFocusWithin` with `:focus-within` — I like it as a separate matcher, but I'd keep it out of this PR to keep the fix focused. Worth its own issue: `:focus-within` also throws in the `node` test project here (`Failed to execute 'contains' on 'Node'`), so it'd need its own jsdom-safe implementation (likely a similar walk checking containment). Happy to open that follow-up if there's interest.

julienw:
Ah right, I'm not using that in jsdom indeed.

